PR #1, opened 2026-08-09
Add ADR-0001 and phase-by-phase implementation plans

karia:
Design and planning only. No application code yet.

- `docs/adr/0001-application-architecture.md` — stack, domain model skeleton (Person/User split, `PlaySession` naming), visibility rule, scope boundary against `yuno04-k3s`
- `docs/plans/` — five phases. Phase 0 is sequential; 1 and 2 run in parallel after it, then 3 and 4

Infrastructure (PostgreSQL, MinIO, Ingress, Cloudflare Tunnel) gets its own ADR in `yuno04-k3s`.


docs/adr/0001-application-architecture.md
```diff
@@ -0,0 +1,119 @@
...
+
+セッションを表すモデルは `PlaySession` と命名する。
+`Session` は Rails のセッション機構と名前が衝突するため使わない。
+
+### 可視性の判定
+
+閲覧者の Person が次のいずれかを満たすとき、その PlaySession を閲覧できる。
+
+- その PlaySession の参加者である
+- その PlaySession の参加者の誰かと、同じグループに所属している
```
karia, line 63: [should-fix] **The "shares a group with any participant" rule can disclose a session to an unrelated group.**

Groups are many-per-person (Phase 2 `GroupMembership`, Phase 4 "グループ所属"). If participant A belongs to both group X and group Y, and A plays a session with group-Y people only, then every member of group X also passes this condition — they share a group with a participant. The wider A's group membership, the wider the leak, and it grows silently every time an admin adds someone to a group.

The original requirement ("visible to people who played or people in the same group") is ambiguous about *whose* group. The alternative reading — the `PlaySession` itself belongs to a group, and visibility is participant OR member-of-that-group — has a bounded blast radius and is easier to explain to users. This is a decision with lasting consequences and it drives the schema (a `group_id` on `play_sessions`), so it should be settled in the ADR before Phase 3 starts, not discovered during implementation.

docs/adr/0001-application-architecture.md
```diff
@@ -0,0 +1,119 @@
...
+
+管理者はすべての PlaySession を閲覧できる。
+
+この判定は Pundit の `Scope` に 1 箇所だけ書き、セッション一覧、セッション詳細、シナリオ詳細に埋め込むセッション履歴のすべてが同じものを使う。
+判定が複数箇所に分かれると、片方だけ直して情報が漏れる形になる。
+
+シナリオの準備情報は、ログイン済みかつ本人がネタバレ防止ボタンを押した場合に限り表示する。
```
karia: [should-fix] **Contradicts Phase 4 on who can open the spoiler guard.**

Here the gate is "ログイン済みかつ本人がネタバレ防止ボタンを押した". `docs/plans/2026-08-09-phase4-profile-and-favorites.md:17` states "準備情報がログイン必須かつ **Person 紐づけ済み** でないと見られない", which is strictly narrower — a logged-in but unlinked user passes the ADR rule and fails the Phase 4 rule.

The requirement only says "logged-in users who pressed the spoiler-guard button", i.e. the ADR reading. But the preparation info sits on the *public* scenario detail page, so making it Person-gated is a defensible extra restriction — it just has to be one decision, stated once, in the ADR. As written, whichever developer picks up Phase 4 will silently choose.

docs/adr/0001-application-architecture.md
```diff
@@ -0,0 +1,119 @@
...
+| Devise の導入 | パスワードと確認メールの機構を使わないため、設定と依存だけが残る |
+| ActiveAdmin | DSL が Hotwire 側と別系統になり、Pundit との連携も薄い |
+
+## 帰結
+
+k3s に PostgreSQL と MinIO という運用対象が 2 つ増える。
+どちらも `yuno04-k3s` 側で別の ADR を立てて扱う。
+
+公開エリアと非公開エリアが 1 つのプロセスに同居するため、認可の漏れがそのまま情報漏えいになる。
+これに対しては、可視性の判定を 1 箇所に集約したうえで、各エンドポイントの request spec で未ログイン、Person 未紐づけ、グループ外、グループ内、管理者の 5 通りを検証する。
```
karia: [should-fix] **Implementation/verification steps in an ADR, and the case count disagrees with Phase 3.**

Two problems in one sentence. First, "各エンドポイントの request spec で ... 5 通りを検証する" is a test plan, which by your own document rules belongs in a plan, not an ADR. Second, the five cases listed here (未ログイン / Person 未紐づけ / グループ外 / グループ内 / 管理者) are not the six in `docs/plans/2026-08-09-phase3-session-area.md:46-52`, which splits "参加者本人" out from "グループ内". Since the ADR is what the plans link to as the source of truth, keep the invariant here (visibility is decided in exactly one Scope) and let Phase 3 own the enumeration of cases.

docs/adr/0001-application-architecture.md
```diff
@@ -0,0 +1,119 @@
...
+
+シナリオの準備情報は、ログイン済みかつ本人がネタバレ防止ボタンを押した場合に限り表示する。
+押した状態は Person とシナリオの組で永続化し、端末をまたいで引き継ぐ。
+
+### 責務の分界
+
+このリポジトリが持つのは、アプリケーションのコード、Dockerfile、CI、ヘルスチェック、そして必要とする環境変数の仕様までとする。
+
+PostgreSQL、MinIO、Ingress、Cloudflare Tunnel の構成は `yuno04-k3s` リポジトリで管理し、ADR もそちらに置く。
+このリポジトリは public であり、クラスタ側の識別子や内部ホスト名を持ち込まない。
```
karia: [should-fix] **This rule is broken by the documents that state it.**

"クラスタ側の識別子や内部ホスト名を持ち込まない" is the right policy for a public repo, but `yuno04` — a specific host/cluster name — appears in this same ADR at line 23 and 77, in `README.md:13`, in `docs/plans/README.md:32-33`, and in `docs/plans/2026-08-09-phase0-foundation.md:32,47`. It also names a presumably private sibling repo `yuno04-k3s`.

Not a leak of credentials, so not blocking, but it is a concrete cluster identifier published to the world and it directly contradicts the stated rule. Either drop the rule, or refer to the target neutrally ("the k3s cluster" / "the infrastructure repository") and keep the mapping out of this repo.

docs/plans/2026-08-09-phase1-scenario-public-area.md
```diff
@@ -0,0 +1,98 @@
...
+- ジャケット画像: Active Storage の添付
+- `StreamLink`: おすすめの配信リンク。複数持つ
+
+各列の値の癖を、スキーマ側で受け止める。
+
+- 「制限なし」の PL人数は最小と最大の両方を NULL にし、補足を `player_count_note` に置く
+- 「1人推奨」のように人数と含みが混ざる行があるため、数値と補足を別の列に分ける
+- 「30分～60分」があるため、目安時間は分単位で持つ
+- オススメ度には空欄と「回したことない」の 2 種類がある。前者は `recommendation` の NULL、後者は `Scenario#run_by_owner` の false で表し、両者を区別する
+- キャラシ提出期限は、前日、前々日、1 週間前、不要、備考参照の 5 つで観測値をすべて覆える。列挙型にする
```
karia: [should-fix] **The enum does not cover all observed values — `"-"` is missing.**

The observed values for キャラシ提出期限 are: セッション前日 / セッション前々日 / セッション1週間前 / 事前準備不要 / メモ参照 / `"-"` — six distinct values, plus blank. This line claims five members "で観測値をすべて覆える", which is not true unless `"-"` is deliberately normalized (to NULL? to 事前準備不要? they mean different things — "unspecified" vs "no deadline").

Since the claim of full coverage is what justifies choosing an enum over free text, say explicitly what `"-"` and blank map to. Getting this wrong means re-migrating after the 60 rows are already hand-entered.

docs/plans/2026-08-09-phase1-scenario-public-area.md
```diff
@@ -0,0 +1,98 @@
...
+- `Scenario#preparation_note`: シナリオ準備情報。Phase 4 でネタバレ防止ボタンの背後に隠す
+- ジャケット画像: Active Storage の添付
+- `StreamLink`: おすすめの配信リンク。複数持つ
+
+各列の値の癖を、スキーマ側で受け止める。
+
+- 「制限なし」の PL人数は最小と最大の両方を NULL にし、補足を `player_count_note` に置く
+- 「1人推奨」のように人数と含みが混ざる行があるため、数値と補足を別の列に分ける
+- 「30分～60分」があるため、目安時間は分単位で持つ
+- オススメ度には空欄と「回したことない」の 2 種類がある。前者は `recommendation` の NULL、後者は `Scenario#run_by_owner` の false で表し、両者を区別する
```
karia: [should-fix] **`run_by_owner` becomes a second source of truth once Phase 3 lands, and it is not in the data-model table.**

Two issues:

1. From Phase 3 onward, "has this scenario been run?" is derivable from whether the scenario has any `PlaySession`. A stored `run_by_owner` flag can and will disagree with the session history shown on the same page. If the flag really means "the owner has never GM'd it", note that the ADR says roles are multi-person from day one (line 52) — "owner" is not a modelled concept, so the column has no stable subject.
2. The flag is introduced only in this bullet. It is absent from the column-mapping table above (lines 13-24) and from the "スプレッドシートにない項目" list (lines 28-31), so the migration list at line 44 will not obviously include it.

A cheaper encoding that keeps the two states distinct without inventing an entity: make `recommendation` nullable *and* add an explicit "未プレイ" enum/state alongside the 1-5 rating, or simply derive it from session history in Phase 3 and drop the flag.

docs/plans/2026-08-09-phase1-scenario-public-area.md
```diff
@@ -0,0 +1,98 @@
...
+- キャラシ提出期限は、前日、前々日、1 週間前、不要、備考参照の 5 つで観測値をすべて覆える。列挙型にする
+- 販売サイトの URL には、URL でない値（購入方法の説明）が入る行がある。`PurchaseLink#url` は NULL を許し、ラベルだけの登録を通す
+
+## 作成するファイル
+
+- マイグレーション: `game_systems`、`authors`、`scenarios`、`scenario_game_systems`、`scenario_authors`、`purchase_links`、`stream_links`
+- モデル: 上記に対応する 7 クラス
+- `app/controllers/scenarios_controller.rb`
+- `app/views/scenarios/index.html.erb`、`show.html.erb` と部分テンプレート
+- `app/policies/scenario_policy.rb`
```
karia: [blocking] **Phase 1 creates a Pundit policy, but Pundit's base classes and `pundit_user` only exist in Phase 2.**

`ScenarioPolicy` needs `ApplicationPolicy` to inherit from, and Pundit needs a `pundit_user` / `current_user` to resolve — both are Phase 2 deliverables (`docs/plans/2026-08-09-phase2-authentication.md:29,31`). Phase 1 declares its only dependency as Phase 0, and Phase 0's file list contains no Pundit setup at all. Run in parallel as the README instructs, Phase 1 either blocks on Phase 2 or invents its own `ApplicationPolicy` that Phase 2 then overwrites.

This is a third Phase 1/Phase 2 merge point, and `docs/plans/README.md:18` says there are exactly two. Cleanest fix: move Pundit installation, `ApplicationPolicy`, and a `current_user` that always returns nil into Phase 0, exactly like the analytics-stub pattern you already use for the other merge points.

docs/plans/2026-08-09-phase1-scenario-public-area.md
```diff
@@ -0,0 +1,98 @@
...
+- 詳細ページの HTML に JSON-LD が含まれ、構造化データテストツールで警告が出ない
+- `/sitemap.xml` が生成され、公開シナリオの URL を列挙している
+- Lighthouse の SEO スコアが 90 以上になる
+- 準備情報が、この時点ではどのユーザーにも表示されない
+
+## 詰まりそうな箇所
+
+**Avo の認可は Phase 2 の Pundit に依存する**。
+Phase 1 の時点では管理者の概念がまだない。
+Avo のルーティングを本番環境では読み込まない形にしておき、Phase 2 で管理者に開く。
```
karia: [should-fix] **Phase 1 cannot meet its own goal if Avo is disabled in production.**

The stated goal (line 3) is "あわせて管理画面からシナリオを登録できるようにする", and line 97 plans to hand-enter the 60-odd existing rows through Avo rather than write an importer. But this sticking point disables the Avo routes in production until Phase 2. The result is a Phase 1 that ships public scenario pages with no way to put data behind them in production — and the SEO verification steps at lines 76-77 (sitemap listing scenarios, Lighthouse ≥ 90) are hard to evaluate against an empty catalogue.

Either move data entry into Phase 2, or gate Avo behind something Phase 1 can actually provide (e.g. HTTP basic auth from an env var) and swap it for Pundit in Phase 2. Also worth noting: this file lists `app/avo/resources/` but nothing installs Avo itself (initializer, `mount Avo::Engine`); that belongs in Phase 0 or explicitly here.

docs/plans/2026-08-09-phase2-authentication.md
```diff
@@ -0,0 +1,85 @@
...
+11. 解析タグと広告タグの描画判定を、ログイン状態を見る実装に差し替える
+
+## 検証手順
+
+- `bin/rspec` がすべて成功する
+- Google でログインでき、ログアウトできる
+- 初回ログインのユーザーが Person 未紐づけで作られる
+- Person 未紐づけのユーザーが、ログイン必須のパスにアクセスすると弾かれる
+- 管理者が Person を作り、`User` に紐づけると、そのユーザーがログイン必須エリアに入れる
+- 管理者以外が `/avo` にアクセスすると 404 または 403 を返す
```
karia: [should-fix] **This acceptance criterion contradicts the requirement that GMs manage content.**

"管理者以外が `/avo` にアクセスすると 404 または 403 を返す" locks out the GM role, but GM-role users are supposed to manage scenarios and sessions through the admin screen — as step 8 (line 45) itself hedges ("GM に開く範囲はリソース単位で決める") and as `docs/plans/2026-08-09-phase3-session-area.md:57` assumes outright ("GM が登録できるようにする").

As written, Phase 3 would have to reopen a decision Phase 2 declared done. Restate the criterion per resource, e.g. "a GM can reach `/avo` and see Scenario/PlaySession/Participation but not Person/User/Group; a plain player gets 404/403".

docs/plans/2026-08-09-phase2-authentication.md
```diff
@@ -0,0 +1,85 @@
...
+
+**OAuth クライアントの登録**。
+リダイレクト URI は開発と本番で別に登録する。
+クライアント ID とシークレットは環境変数で渡し、リポジトリに置かない。
+
+**CSRF 対策**。
+omniauth-rails_csrf_protection を入れ、認証の開始を POST に限定する。
+GET のままだと、外部サイトからログインを誘発できる。
+
+**権限の置き場所**。
```
karia, line 72: [optional] **Rationale in a plan, duplicated from the ADR.**

"`User` に権限を持たせると、アカウントを差し替えたときに権限が失われる" is a justification for a decision the ADR already records (line 52, plus the account-swap consequence at line 119). Per your document rules, plans link to the ADR rather than re-argue it — and duplicated rationale is what drifts first when the ADR is revised.

karia (commented):
Third-party review of the ADR and the five phase plans against the original requirements. Overall the split is sound (Person/User, `PlaySession` naming, single Pundit Scope for visibility, the Phase 1/2 merge points), and every requirement has a home somewhere. The findings below are concentrated in four areas: the group-based visibility rule can over-disclose, Phase 1 and Phase 2 collide on Pundit and on Avo in a way the README's "two merge points" does not cover, a few ADR/plan statements contradict each other (spoiler gate, GM access to the admin screen, 5 vs 6 authorization cases), and one observed spreadsheet value is not covered by the proposed enum.

docs/plans/2026-08-09-phase3-session-area.md
```diff
@@ -0,0 +1,90 @@
...
+## データモデル
+
+- `PlaySession`: `scenario_id`、`played_at`（NULL 可）、`status`（予定 / 実施済み）、`recording_url`（NULL 可）
+- `Participation`: `play_session_id`、`person_id`、`role`（GM / PL / サブキーパー）、`character_name`（NULL 可）、`character_sheet_url`（NULL 可）
+
+キャラクターシートのリンクは参加者ごとに持つ。
+GM とサブキーパーには存在しないことがあるため NULL を許す。
+
+日時は未定を許す。
+予定のセッションでは日付だけ決まっていて時刻が未定の場合があるため、日付と時刻を分けて持つか、日時を NULL 可にしたうえで補足を持たせるかを実装前に決める。
```
karia: [should-fix] **The data-model section both declares the column and says its shape is undecided.**

Line 11 states `played_at`（NULL 可）; seven lines later the plan says it is still open whether to split date and time or keep a nullable timestamp plus a note. Since "a future plan whose time is not fixed yet" is an explicit requirement, this is a real modelling choice with migration consequences, not a detail — decide it before Phase 3 opens, and if it has lasting consequences put it in the ADR.

Related: `status`（予定 / 実施済み）overlaps with `played_at` vs. now. Two representations of the same fact drift; if `status` exists to cover "scheduled but the date passed and it did not happen", say so, otherwise derive it.

docs/plans/2026-08-09-phase3-session-area.md
```diff
@@ -0,0 +1,90 @@
...
+セッション一覧、セッション詳細、シナリオ詳細のセッション履歴は、すべてこの Scope を通す。
+
+## 作成するファイル
+
+- マイグレーション: `play_sessions`、`participations`
+- モデル: `PlaySession`、`Participation`
+- `app/policies/play_session_policy.rb`
+- `app/controllers/play_sessions_controller.rb`
+- `app/views/play_sessions/index.html.erb`、`show.html.erb`
+- `app/views/scenarios/_play_session_history.html.erb`
```
karia: [should-fix] **Phase 3 and Phase 4 both edit the scenario detail page, and they are scheduled in parallel.**

This partial goes into `app/views/scenarios/show.html.erb`, and so does `app/views/scenarios/_preparation_note.html.erb` from `docs/plans/2026-08-09-phase4-profile-and-favorites.md:39`. `docs/plans/README.md:16` says Phase 3 and Phase 4 run in parallel, so two people will edit the same template — the one collision the phase split was designed to avoid.

You already solved this shape of problem for the Phase 1/Phase 2 analytics tags: have Phase 1 render both partials as empty stubs from `show.html.erb`, so Phase 3 and Phase 4 each only fill in their own file. The same applies to Avo's resource registration if it is a single central list rather than per-file.

docs/plans/2026-08-09-phase4-profile-and-favorites.md
```diff
@@ -0,0 +1,88 @@
...
+
+## データモデル
+
+- `PersonAlias`: `person_id`、`name`、`context`（どの Discord サーバでの名前かの補足）。1 人が複数持つ
+- `Favorite`: `person_id`、`scenario_id`
+- `SpoilerReveal`: `person_id`、`scenario_id`。ネタバレ防止ボタンを押した記録
+
+Person の表示名、アイコン、X のアカウント名は Phase 2 で作った `people` に既にある。
+
+`SpoilerReveal` を Person に紐づけるのは、準備情報がログイン必須かつ Person 紐づけ済みでないと見られないためである。
```
karia: [optional] **Decision rationale in a plan.**

These two sentences argue *why* `SpoilerReveal` hangs off `Person` rather than `User`. Per your document rules that reasoning belongs in the ADR (which already establishes the Person/User split); the plan should just list the table and link. Note this also states the stricter Person-linked gate that contradicts ADR line 70 — see the comment there.

docs/plans/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## フェーズと依存
+
+| フェーズ | 内容 | 依存 |
+| --- | --- | --- |
+| [Phase 0](2026-08-09-phase0-foundation.md) | 土台。Rails の雛形から k3s での起動確認まで | なし |
+| [Phase 1](2026-08-09-phase1-scenario-public-area.md) | シナリオ公開エリアと管理画面での登録 | Phase 0 |
+| [Phase 2](2026-08-09-phase2-authentication.md) | 認証、権限、グループ | Phase 0 |
+| [Phase 3](2026-08-09-phase3-session-area.md) | セッションエリアと可視性 | Phase 1、Phase 2 |
+| [Phase 4](2026-08-09-phase4-profile-and-favorites.md) | プロフィール、別名、お気に入り、ネタバレ防止ボタン | Phase 2 |
```
karia: [should-fix] **The dependency table understates Phase 4's dependencies.**

This row says Phase 4 depends on Phase 2 only, but the phase document itself (`2026-08-09-phase4-profile-and-favorites.md:5`) says favorites and the spoiler-guard button also wait on Phase 1 because they reference `Scenario`. Since this table is what someone scheduling parallel work will read, it should say "Phase 2（お気に入りとネタバレ防止は Phase 1 も）". With that correction, only the profile/alias half of Phase 4 can actually start alongside Phase 1.

docs/adr/0001-application-architecture.md
karia: Not changed — deliberate.

The alternative (attaching a group to the `PlaySession`) was presented to the repo owner alongside the derive-from-participants rule before the ADR was written, and they picked derivation. The disclosure property you describe is a consequence of that choice, not a defect in the write-up.

What did change in baa02b9: the consequence is now recorded in the ADR '帰結' section, including that group granularity has to be maintained operationally and that moving visibility onto the session itself is the escape hatch if that stops working.

docs/adr/0001-application-architecture.md
karia: Fixed in baa02b9.

ADR now reads 「Person に紐づいたユーザーがネタバレ防止ボタンを押した場合に限り表示する」, matching Phase 4. Since `SpoilerReveal` is keyed by `person_id`, Person linkage is the real precondition and the ADR was the wrong one.

docs/adr/0001-application-architecture.md
karia: Fixed in baa02b9.

The enumeration of test cases is gone from the ADR; the sentence now states only the consequence (co-located public and private areas mean an authorization gap is a disclosure, which is why the check is centralized). The 6 cases stay in Phase 3, where they belong.

docs/adr/0001-application-architecture.md
karia: Fixed in baa02b9.

Reworded to draw the line where it actually is: credentials and cluster-side resource identifiers (tunnel IDs, zone IDs) stay out; the cluster name and the repo name may be referenced.

For the record, both `yuno04` and `karia/yuno04-k3s` are already published in the public `dot-files` repo (`.claude/skills/publishing-k3s-apps/SKILL.md`), so nothing new was disclosed here. gitleaks over this branch was clean.

docs/plans/2026-08-09-phase1-scenario-public-area.md
karia: Fixed in baa02b9.

The line now says the 5 values are the enum and that `"-"` and blank are NULL, displayed as 未設定. The enum column is nullable.

docs/plans/2026-08-09-phase1-scenario-public-area.md
karia: Fixed in baa02b9.

Renamed to `gm_experienced` and moved into the column-mapping table alongside `recommendation`, so the migration list picks it up. The three states are now spelled out: starred rows get 1–5; 「回したことない」 sets `gm_experienced` false with `recommendation` NULL; blank rows leave `gm_experienced` true with `recommendation` NULL.

On the redundancy with `PlaySession`: kept deliberately. Deriving it requires every past session to have been entered, and Phase 1 ships before `PlaySession` exists.

docs/plans/2026-08-09-phase1-scenario-public-area.md
karia: Fixed in baa02b9.

Pundit and Avo both moved into Phase 0, which now ships a deny-by-default `ApplicationPolicy` and a `current_user` that returns nil. Phase 2 swaps the contents, the same stub pattern already used for the analytics gate. `ScenarioPolicy` in Phase 1 now says it inherits the Phase 0 base.

`docs/plans/README.md` keeps the merge-point count at two, since Pundit and Avo are no longer a merge point — Phase 0 owns them.

docs/plans/2026-08-09-phase1-scenario-public-area.md
karia: Fixed in baa02b9.

Phase 1 now mounts Avo behind HTTP Basic auth with credentials from an env var, and Phase 2 replaces that with the Pundit check. Data entry for the ~60 rows is an explicit step (step 9), and the SEO verification line now says to measure after the data is in. Avo installation is in Phase 0.

docs/plans/2026-08-09-phase2-authentication.md
karia: Fixed in baa02b9.

The criterion now reads 「管理者でも GM でもないユーザーが `/avo` にアクセスすると 404 を返す」, plus a second line checking that GMs can enter Avo but cannot touch the Person and Group resources. Step 8 states the split: admins get every resource, GMs get scenarios and sessions.

docs/plans/2026-08-09-phase2-authentication.md
karia: Not changed.

This sits under 詰まりそうな箇所, which exists to warn the implementer away from a plausible mistake — putting roles on `User` is exactly what someone would do by default. The ADR records the decision; this records the trap. Removing it would cost more than the duplication does.

docs/plans/2026-08-09-phase3-session-area.md
karia: Fixed in baa02b9.

Split into `played_on` (date, nullable) and `started_at` (time, nullable); the hedging sentence is gone and replaced with the reason a single datetime column cannot express 「日付だけ確定」.

`status` stays, and is now explicitly not derived from the date: a past-dated plan that was cancelled is indistinguishable from one that happened if you only look at the date. Added 中止 to the enum for that case.

docs/plans/2026-08-09-phase3-session-area.md
karia: Fixed in baa02b9.

Phase 1 now creates both `_play_session_history.html.erb` and `_preparation_note.html.erb` as empty partials and wires them into `show.html.erb`. Phase 3 and Phase 4 each fill one and neither touches `show.html.erb`. `docs/plans/README.md` states this explicitly so the parallel implementers see it.

docs/plans/2026-08-09-phase4-profile-and-favorites.md
karia: Fixed in baa02b9. Both sentences removed; the ADR already records why `SpoilerReveal` hangs off `Person`.

docs/plans/README.md
karia: Fixed in baa02b9. The row now reads 「Phase 2。お気に入りとネタバレ防止ボタンは Phase 1 も」, matching the phase document.

PR #2, opened 2026-08-09
Phase 0: Rails foundation, MinIO storage, CI and container image

karia:
## What

Rails 8 skeleton through to a container image that answers `/up`. Covers the app-side of [Phase 0](docs/plans/2026-08-09-phase0-foundation.md).

- PostgreSQL config built from discrete env vars, not `DATABASE_URL` — CloudNativePG splits host/user/password across Secret keys
- Rails credentials removed; `SECRET_KEY_BASE` comes from the environment
- Active Storage on MinIO (`force_path_style`), local disk in test
- Pundit and Avo installed as deny-by-default foundations so Phase 1 and Phase 2 do not wait on each other. Avo is not mounted yet
- Migrations dropped from the container entrypoint; a release Job runs them

## Why

Phase 0 is the only sequential phase. Everything here exists so Phase 1 and Phase 2 can start in parallel.

## Verification

- [x] `bin/rspec` — 9 examples, 0 failures
- [x] `prek run --all-files`
- [x] `docker build` + `curl /up` → 200 against the compose PostgreSQL
- [x] Active Storage round trip against MinIO in development

## Related

Not in this PR: CloudNativePG, MinIO and the Ingress live in `yuno04-k3s`. Steps 12–13 of the plan stay open until that lands.


config/initializers/avo.rb
```diff
@@ -0,0 +1,206 @@
...
+  #   show: 'show?',
+  #   edit: 'edit?',
+  #   new: 'new?',
+  #   update: 'update?',
+  #   create: 'create?',
+  #   destroy: 'destroy?',
+  #   search: 'search?',
+  # }
+  # config.raise_error_on_missing_policy = false
+  config.authorization_client = nil
```
karia: [blocking] **`authorization_client = nil` disables Avo authorization entirely — this is the opposite of the deny-by-default foundation the PR describes.**

Avo 4.1.3's default is `:pundit` (`lib/avo/configuration.rb:177`). Setting it to `nil` makes `Avo::Configuration#authorization_enabled?` return false:

```ruby
@authorization_enabled ||= Avo.plugin_manager.installed?("avo-authorization") && !authorization_client.nil?
```

So the moment Phase 1 calls `mount_avo`, every resource is reachable by anyone who gets past the mount guard, with no policy consulted — `ApplicationPolicy` returning false everywhere buys nothing inside Avo.

Worse, note the *first* half of that condition: authorization in Avo 4 requires the `avo-authorization` plugin, which is **not** in `Gemfile.lock` (only `avo (4.1.3)` and `avo-icons (0.2.1)`). Even restoring `:pundit` here will not enable policy checks on the community gem. Phase 1 is going to hit this wall.

This is worth resolving now, in Phase 0, since it is exactly the kind of cross-phase dependency this PR exists to remove. Either add the `avo-authorization` plugin and set `authorization_client = :pundit` (plus `raise_error_on_missing_policy = true` and `explicit_authorization = true`, both currently commented out at lines 38/40), or drop the claim from the PR body / CLAUDE.md line 28 and record that Avo will be protected solely by the mount guard, so Phase 1 designs for that.

Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+RUN apt-get update -qq && \
+    apt-get install --no-install-recommends -y curl libjemalloc2 libvips postgresql-client && \
+    ln -s /usr/lib/$(uname -m)-linux-gnu/libjemalloc.so.2 /usr/local/lib/libjemalloc.so && \
+    rm -rf /var/lib/apt/lists /var/cache/apt/archives
+
+# Set production environment variables and enable jemalloc for reduced memory usage and latency.
+ENV RAILS_ENV="production" \
+    BUNDLE_DEPLOYMENT="1" \
+    BUNDLE_PATH="/usr/local/bundle" \
+    BUNDLE_WITHOUT="development" \
```
karia: [should-fix] **`BUNDLE_WITHOUT="development"` no longer excludes the test gems, because the Gemfile moved them into `group :test`.**

The stock Rails value works because the generated Gemfile puts Capybara and Selenium in `group :development, :test`. This Gemfile has a dedicated `group :test` (Gemfile lines 76-79), and gems in `:development, :test` are not excluded either since they are still in a non-excluded group. Result: `capybara`, `selenium-webdriver`, `websocket`, `rspec-rails`, `factory_bot_rails`, `dotenv-rails`, `brakeman`, `bundler-audit`, `rubocop-rails-omakase` and `debug` are all installed into the runtime image.

They aren't `require`d at runtime (`Rails.groups` in production is `[:default, :production]`), so this is image size and attack surface rather than a live bug — but shipping the `debug` gem and a browser automation stack in the production image is not what you want.

```
BUNDLE_WITHOUT="development:test"
```

config/storage.yml
```diff
@@ -0,0 +1,13 @@
...
+  root: <%= Rails.root.join("tmp/storage") %>
+
+# MinIO はバケット名をサブドメインで解決しないため force_path_style が要る。
+minio:
+  service: S3
+  endpoint: <%= ENV.fetch("S3_ENDPOINT", "http://localhost:9000") %>
+  access_key_id: <%= ENV.fetch("S3_ACCESS_KEY_ID", "minioadmin") %>
+  secret_access_key: <%= ENV.fetch("S3_SECRET_ACCESS_KEY", "minioadmin") %>
+  region: <%= ENV.fetch("S3_REGION", "us-east-1") %>
+  bucket: <%= ENV.fetch("S3_BUCKET", "trpg-catalog-development") %>
```
karia, line 12: [should-fix] **Production silently inherits the development MinIO defaults if any `S3_*` variable is missing.**

`production.rb:25` sets `config.active_storage.service = :minio`, and this is the same `minio:` block development uses. A Secret key typo or a missed `env` entry in the Deployment doesn't fail the boot — the app comes up healthy (`/up` only checks Postgres) and then writes to `trpg-catalog-development` at `http://localhost:9000` with `minioadmin`/`minioadmin`. That failure surfaces as a 500 on the first jacket upload, in production, well after the deploy looks green.

Suggest either splitting a `minio_production:` entry that uses bare `ENV.fetch("S3_BUCKET")` (no default) so a missing variable is a boot-time crash, or dropping the defaults here and putting them only in `.env.example` / `compose.yaml`. Fail fast is cheap; the alternative is silent misconfiguration.

app/controllers/application_controller.rb
```diff
@@ -0,0 +1,10 @@
+class ApplicationController < ActionController::Base
+  include Authentication
+  include Pundit::Authorization
```
karia, line 3: [should-fix] **Pundit is included but nothing enforces that `authorize` was ever called, and `Pundit::NotAuthorizedError` is unhandled.**

ADR-0001 says the public and private areas share one process and "認可の漏れがそのまま情報漏えいになる". Deny-by-default policies only help on the code paths that actually call `authorize` / `policy_scope`; a Phase 1 controller action that forgets the call renders fine, with no policy consulted. The safety net for that is:

```ruby
after_action :verify_authorized
after_action :verify_policy_scoped, only: :index
rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
```

with the public scenario controllers opting out explicitly via `skip_after_action`. Adding this in Phase 0 is what makes "deny by default" a property of the app rather than of each future controller.

Separately, without a `rescue_from`, an authorization failure in production is an unhandled 500 rather than a 403/404 — Phase 1 and Phase 2 will both need this, so it belongs on the shared base.

config/environments/production.rb
```diff
@@ -0,0 +1,90 @@
...
+  config.i18n.fallbacks = true
+
+  # Do not dump schema after migrations.
+  config.active_record.dump_schema_after_migration = false
+
+  # Only use :id for inspections in production.
+  config.active_record.attributes_for_inspect = [ :id ]
+
+  # Enable DNS rebinding protection and other `Host` header attacks.
+  # config.hosts = [
```
karia: [should-fix] **Host authorization is left commented out on an internet-exposed app.**

`config.hosts` is only populated by default in the development environment, so production accepts any `Host` header. Behind a Cloudflare Tunnel the tunnel itself constrains the hostname reaching the pod, but this is one Ingress or port-forward away from being the app's only defence, and it is the standard mitigation for absolute-URL poisoning (password/confirmation links, cache poisoning) once Phase 2 adds OmniAuth callbacks and mailers.

Suggest enabling it, and note that you must keep the `exclude` below it or the k3s liveness/readiness probes — which hit the pod IP, not `trpg-catalog.side2.net` — will start getting 403:

```ruby
config.hosts = [ ENV.fetch("APP_HOST", "trpg-catalog.side2.net") ]
config.host_authorization = { exclude: ->(request) { request.path == "/up" } }
```

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,29 @@
+# Be sure to restart your server when you modify this file.
+
+# Define an application-wide content security policy.
+# See the Securing Rails Applications Guide for more information:
+# https://guides.rubyonrails.org/security.html#content-security-policy-header
+
+# Rails.application.configure do
```
karia: [should-fix] **CSP is entirely commented out, and `csp_meta_tag` in the layout implies otherwise.**

`app/views/layouts/application.html.erb:10` renders `csp_meta_tag`, which emits nothing while no policy is configured. For a public, SSR, internet-facing app that will render user/admin-authored text (scenario notes, session records) this is the cheapest XSS mitigation available, and ADR-0001 also plans to inject Google Analytics/AdSense tags for anonymous visitors — deciding the policy *after* those third-party scripts land is much harder than deciding it now.

Phase 0 is the right moment: enable a policy with the nonce generator (importmap + Stimulus need `script_src :self` plus nonces), and consider shipping it as `content_security_policy_report_only = true` first so it can't break Phase 1/2 while they build.

config/ci.rb
```diff
@@ -0,0 +1,20 @@
+# Run using bin/ci
+
+CI.run do
+  step "Setup", "bin/setup --skip-server"
+
+  step "Style: Ruby", "bin/rubocop"
+
+  step "Security: Gem audit", "bin/bundler-audit"
+  step "Security: Importmap vulnerability audit", "bin/importmap audit"
+  step "Security: Brakeman code analysis", "bin/brakeman --quiet --no-pager --exit-on-warn --exit-on-error"
```
karia, line 11: [should-fix] **`bin/ci` never runs the test suite.**

The generated `config/ci.rb` carries a `step "Tests: Rails", "bin/rails test"`; it was removed here (note the blank gap at lines 11-12) and not replaced with an RSpec equivalent. So the local "run everything" entry point that `bin/ci` exists to be runs setup, RuboCop, bundler-audit, importmap audit and Brakeman — and reports success without executing a single spec. GitHub Actions does run `bin/rspec` (`ci.yml:104`), so this only misleads locally, but that is precisely who `bin/ci` is for.

```ruby
step "Tests: RSpec", "bin/rspec"
```

While there: `bundle exec erb_lint --lint-all` runs in CI but not here either.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    branches: [ main ]
+
+jobs:
```
karia, line 11: [should-fix] **No `permissions` block — this workflow runs with the repository's default `GITHUB_TOKEN` scopes.**

`build.yml` correctly declares job-level permissions; `ci.yml` declares none, so every job inherits the repo default, which is read/write for repositories created before the default changed. On a public repo, CI runs against `pull_request` from any fork, and a compromised dependency in `bundler-cache: true` (which executes `bundle install`, running gem extension builds) would hold a writable token.

Add at the top level:

```yaml
permissions:
  contents: read
```

compose.yaml
```diff
@@ -0,0 +1,51 @@
...
+  minio-init:
+    image: minio/mc:latest
+    depends_on:
+      minio:
+        condition: service_healthy
+    entrypoint: >
+      sh -c "
+      mc alias set local http://minio:9000 minioadmin minioadmin &&
+      mc mb --ignore-existing local/trpg-catalog-development &&
+      mc anonymous set download local/trpg-catalog-development
```
karia: [should-fix] **`mc anonymous set download` makes the development bucket world-readable, which is the configuration the proxy-mode decision exists to avoid.**

Jacket images are to be served through Active Storage proxy mode rather than exposing MinIO. With the dev bucket anonymously downloadable, a direct-URL bug in Phase 1 renders images perfectly on the developer's machine and only fails against the real cluster — the local environment silently disagrees with production on the one property that matters here.

Dropping this line costs nothing (Active Storage authenticates with the root credentials for both read and write) and makes development match the deployed behaviour.

config/environments/production.rb
```diff
@@ -0,0 +1,90 @@
...
+  config.action_controller.perform_caching = true
+
+  # Cache assets for far-future expiry since they are all digest stamped.
+  config.public_file_server.headers = { "cache-control" => "public, max-age=#{1.year.to_i}" }
+
+  # Enable serving of images, stylesheets, and JavaScripts from an asset server.
+  # config.asset_host = "http://assets.example.com"
+
+  # Store uploaded files on the local file system (see config/storage.yml for options).
+  config.active_storage.service = :minio
```
karia, line 25: [should-fix] **Missing `resolve_model_to_route` seam for the already-decided proxy delivery mode.**

Jacket images are to be served via Active Storage proxy, not by exposing MinIO — but nothing here or in `application.rb` sets:

```ruby
config.active_storage.resolve_model_to_route = :rails_storage_proxy
```

Without it, `url_for(scenario.jacket)` and every `image_tag @scenario.jacket` in Phase 1 resolves to `rails_blob_url`, i.e. a redirect to a presigned MinIO URL that is not reachable from the internet. That is a broken-image bug that only appears once deployed, and it will be found by whoever is furthest into Phase 1 rather than here. Setting it in Phase 0 costs one line and removes the trap.

(Also worth setting `config.active_storage.draw_routes` intentionally, and pinning a `variant_processor`/`content_types_allowed_inline` policy while you are in here.)

config/database.yml
```diff
@@ -0,0 +1,35 @@
...
+  <<: *default
+  database: trpg_catalog_test
+
+production:
+  primary:
+    <<: *default
+    database: <%= ENV.fetch("DATABASE_NAME", "trpg_catalog") %>
+  cache:
+    <<: *default
+    database: <%= ENV.fetch("CACHE_DATABASE_NAME", "trpg_catalog_cache") %>
```
karia, line 26: [should-fix] **The three extra production databases need cluster-side provisioning that the release Job cannot do on its own.**

Splitting cache/queue/cable off the primary is a sound call, but the release Job's `db:prepare` will `CREATE DATABASE trpg_catalog_cache` (and `_queue`, `_cable`) using the credentials in `DATABASE_USERNAME`. CloudNativePG's `app` role is created with `LOGIN` only — it does not get `CREATEDB` — so unless `yuno04-k3s` either declares the three databases (CNPG `Database` resources) or grants `createdb: true` via `managed.roles`, the first release Job fails with a permission error and the deploy never reaches a Ready pod.

The PR body defers steps 12-13 to the other repo, which is fine; this is a request to write the requirement down somewhere the k3s work will see it (the `.env.example` comment block or CLAUDE.md), since the constraint is not discoverable from the app side.

spec/models/active_storage_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+RSpec.describe "Active Storage" do
+  it "stores a blob and reads the same bytes back" do
+    blob = ActiveStorage::Blob.create_and_upload!(
```
karia, line 5: [optional] **This spec passes regardless of whether the MinIO configuration this PR adds is correct.**

`test.rb:32` sets `config.active_storage.service = :test`, i.e. the Disk service — so the spec exercises Active Storage's model layer and the local filesystem, never `config/storage.yml`'s `minio:` entry, the `force_path_style` setting the plan calls out as a known trap, or the `aws-sdk-s3` code path. Break the endpoint, the bucket, or `force_path_style` and this stays green.

Using Disk in CI is the right trade-off (the plan says as much), but the spec name and the plan's step 5 / verification list both claim MinIO coverage that isn't here. Either rename it to describe what it actually covers, or add a spec tagged to run only when `S3_ENDPOINT` is reachable so `docker compose up` + `bin/rspec` does verify the real path locally.

spec/requests/avo_mount_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "rails_helper"
+
+RSpec.describe "Avo mount point" do
+  it "is not reachable until Phase 1 puts a guard in front of it" do
+    get "/avo"
+
```
karia: [optional] **This assertion holds for any unrouted path, so it doesn't test what its description claims.**

`get "/zzz"` produces the same 404. The spec cannot distinguish "Avo is installed but deliberately unmounted" from "Avo isn't installed at all" or "routes.rb is empty" — and it will not catch the risk it is named after, which is Avo being mounted *without* a guard. Strengthening it slightly would help, e.g. also assert `defined?(Avo::Engine)` so the spec fails if Avo disappears, and assert `Avo.configuration.root_path == "/avo"` so the path under test is derived rather than hardcoded.

Related: given the finding on `config/initializers/avo.rb:39`, a spec asserting `Avo.configuration.authorization_enabled?` is what would actually protect this boundary once Phase 1 mounts the engine.

app/controllers/health_controller.rb
```diff
@@ -0,0 +1,10 @@
+# Kubernetes の probe が User-Agent を持たないため allow_browser を通さない。
+class HealthController < ActionController::Base
+  # 接続は遅延生成されるため active? では判定できない。verify! で実際につなぐ。
+  def show
+    ActiveRecord::Base.connection.verify!
+    head :ok
+  rescue StandardError
```
karia: [optional] **The bare rescue swallows the reason the health check failed.**

When this starts returning 503 in the cluster, the pod gets killed and restarted by the liveness probe with nothing in the logs explaining why — and `config.silence_healthcheck_path = "/up"` (production.rb:44) suppresses the request log line too, so there is no trace at all. A one-line `Rails.logger.error` in the rescue turns an opaque CrashLoopBackOff into a readable one.

Also consider narrowing to `rescue ActiveRecord::ActiveRecordError, PG::Error` — as written, a `NoMethodError` from a future edit to this action is reported as "database unreachable".

.github/workflows/build.yml
```diff
@@ -0,0 +1,62 @@
+name: Build
+
+on:
+  push:
+    branches: [ main ]
```
karia, line 5: [optional] **The image is only ever built after merge, so Dockerfile regressions land on `main`.**

Nothing in `ci.yml` builds the container, and this workflow triggers on `push` to `main` and tags. A change that breaks `bundle install`, `assets:precompile` or the `SECRET_KEY_BASE_DUMMY` path passes all PR checks and only fails once it is already merged — which for a single-node deployment means `main` is briefly not deployable.

Cheap fix: add a job to `ci.yml` using `docker/build-push-action` with `push: false` and `cache-from: type=gha`, so PRs prove the image still builds while the publishing path stays main-only.

karia (commented):
Independent review of the Phase 0 foundation. Overall this is a clean, well-scoped skeleton and the documented deviations (discrete DB env vars, no credentials, `scope.none`, separate Solid databases) are all defensible. The main things I'd fix before this becomes the base for two parallel phases: Avo authorization is actually switched **off**, not deny-by-default; `BUNDLE_WITHOUT` no longer excludes the test gems the Gemfile moved into `group :test`; and production silently inherits the development MinIO defaults. Details inline, tagged blocking / should-fix / optional.

config/initializers/avo.rb
karia: Correct, and it goes further than the config line. Confirmed against the installed gem: `avo-4.1.3/lib/avo/configuration.rb:312` gates on `plugin_manager.installed?("avo-authorization")`, and `avo-authorization` is not on public RubyGems (404) — it ships only with Avo's paid plans. Avo 4 OSS cannot do Pundit authorization at all.

Resolved by dropping Avo entirely (5708af9). The premise behind picking it was wrong.

The owner reframed the requirement while deciding: the editing screens are used by GMs, who are fellow players rather than operators, so a separate admin UI splits the experience for the people who use it most. Scenario and session editing are both nested forms (multiple authors, systems, purchase links, stream links; per-participant role, character name and sheet URL), which is where admin gems are weakest. And the admin/GM split has to be hand-written under any of them.

So: ordinary Rails screens under Pundit, with `ransack` for list filtering and `pagy` for pagination. ADR-0001 and all four phase plans updated in the same commit.

Dockerfile
karia: Confirmed and fixed in 02f827c (`BUNDLE_WITHOUT="development:test"`).

Verified by inspecting the built image rather than trusting the Gemfile — `ls /usr/local/bundle/ruby/*/gems` showed capybara-3.40.0, selenium-webdriver-4.46.0, rspec-rails-8.0.4, brakeman-8.0.5, debug-1.11.1. After the fix the same command matches zero of them.

config/storage.yml
karia: Fixed in 02f827c, though not the way you suggested.

Making the production entries `ENV.fetch` without a default does not work: Rails parses the whole of `config/storage.yml` at boot, so a missing production key would break development too.

Instead there is now `config/initializers/required_env.rb`, which fails the boot in production when any of `SECRET_KEY_BASE`, the four `DATABASE_*` vars, or the four `S3_*` vars is blank. That covers the same hole and the ones next to it. `SECRET_KEY_BASE_DUMMY` is excluded so `assets:precompile` still works during the image build — verified by rebuilding.

app/controllers/application_controller.rb
karia: Fixed in 02f827c. `after_action :verify_authorized` is on in every environment, not just development — ADR-0001 treats an authorization gap as a disclosure, so failing closed everywhere is the right side to err on. Controllers that genuinely need no check will `skip_after_action :verify_authorized` explicitly.

`rescue_from Pundit::NotAuthorizedError` now answers 404 rather than 403, so a denied session detail does not confirm that the session exists.

config/environments/production.rb
karia: Fixed in 02f827c. `config.hosts = [ ENV.fetch("APP_HOST", "trpg-catalog.side2.net") ]` with the `/up` exclude you flagged.

Verified on the cluster: through Traefik with `Host: trpg-catalog.side2.net` the app answers, and `/up` answers 200 for the kubelet probe, which sends no Host header.

config/initializers/content_security_policy.rb
karia: Fixed in 02f827c. A restrictive baseline is now enforced (`default_src :self`, `object_src :none`, `frame_ancestors :none`, `base_uri`/`form_action` pinned), with nonce generation for script and style.

You are right that this only gets harder later. Phase 1 adds GA and AdSense, and those origins now have to be added deliberately rather than arriving under a wide-open policy.

config/ci.rb
karia: Fixed in 02f827c. `bin/ci` now runs `bin/rspec` and `erb_lint` alongside the existing steps, so local and GitHub runs check the same things.

.github/workflows/ci.yml
karia: Fixed in 02f827c. `permissions: contents: read` at workflow level.

`build.yml` keeps its wider grants (packages, attestations, id-token) because it pushes to GHCR and signs provenance, and it only runs on `main`, never on fork PRs.

compose.yaml
karia: Fixed in 02f827c. The `mc anonymous set download` line is gone; the development bucket is now private, matching the proxy-mode decision.

config/environments/production.rb
karia: Fixed in 02f827c. `config.active_storage.resolve_model_to_route = :rails_storage_proxy` in production.

Confirming the surrounding decision, since it was made after this PR was opened: MinIO stays inside the cluster with no Ingress, and images are served from the app host through Cloudflare. ADR-0001 now records this.

config/database.yml
karia: Correct that the app role cannot `CREATE DATABASE`, and this was already handled cluster-side while the review was running.

`karia/yuno04-k3s` now carries `apps/trpg-catalog/postgres.yaml` with a CloudNativePG `Cluster` plus three `Database` resources (`trpg_catalog_cache`, `trpg_catalog_queue`, `trpg_catalog_cable`), each owned by the app role. The operator creates them; `db:prepare` only migrates into them.

Verified end to end on the cluster: the release Job ran `db:prepare` to completion against all four databases, and `trpg_catalog_cache` came out with its three Solid Cache tables.

spec/models/active_storage_spec.rb
karia: Agreed that the spec does not exercise MinIO, and the plan overstated it.

Not changed here, deliberately. Running the suite against MinIO would mean standing one up in CI for a single example, and Phase 0 explicitly chose Disk in test to keep the job light.

The MinIO path is verified out of band instead, and was re-verified for this branch: an Active Storage round trip in development against the compose MinIO, and the same again inside the cluster after deploying the built image. Worth saying plainly that this is a manual check, not a guarded one.

spec/requests/avo_mount_spec.rb
karia: Moot — the file is gone along with Avo (5708af9). The assertion was weak for exactly the reason you give.

app/controllers/health_controller.rb
karia: Fixed in 02f827c. The rescue now logs the exception class and message before answering 503.

.github/workflows/build.yml
karia: Fixed in 02f827c, in `ci.yml` rather than `build.yml`: an `image` job builds the production image on every PR with `push: false`, sharing the GHA build cache with the release build. `build.yml` stays main-only for the push and attestation.

PR #9, opened 2026-08-09
Phase 1: public scenario list and detail, plus editing screens

karia:
## What

Implements [Phase 1](docs/plans/2026-08-09-phase1-scenario-public-area.md) except the data entry itself.

- Domain models for the spreadsheet's columns, including the three-state recommendation (`recommendation` + `gm_experienced`) and nullable ends for player count and duration
- Public list and detail pages, readable without signing in
- Editing screens under `/manage`, gated by HTTP basic auth until Phase 2 replaces it with Pundit
- SEO: meta-tags, JSON-LD on the detail page, sitemap, `robots.txt` disallowing `/manage/`
- Analytics and AdSense partials that render only when signed out; CSP widened for exactly those origins

## Design

Mincho for titles, monospace for the vital statistics so they align down the list — those numbers are what you pick a scenario by. Concealment is the signature: the preparation note renders as a solid ink bar rather than hidden text, which is also what Phase 4's spoiler button and Phase 3's session history will use. Japanese system fonts only, since CSP is `default_src :self`.

## Notable

- `HttpUrlValidator` parses with `URI` instead of a prefix regex. Brakeman flagged the regex, and a newline could smuggle a second scheme past it.
- The preparation note never reaches the client. It is omitted server-side, not hidden with CSS — a spec asserts the string is absent from the body.
- `scenarios/show.html.erb` renders two empty partials, `_play_session_history` and `_preparation_note`. Phase 3 and Phase 4 fill one each and neither touches `show.html.erb`.

## Verification

- [x] `bin/rspec` — 52 examples, 0 failures
- [x] `prek run --all-files`
- [x] List and detail return 200 signed out; preparation note absent from the response body
- [x] `sitemap:refresh:no_ping` produces 4 links
- [x] Basic auth missing → 401 on both read and write under `/manage`

## Not in this PR

Entering the ~60 scenarios from the spreadsheet, and the Lighthouse SEO measurement that depends on it.


app/views/layouts/application.html.erb
```diff
@@ -1,31 +1,41 @@
 <!DOCTYPE html>
-<html>
+<html lang="ja">
   <head>
-    <title><%= content_for(:title) || "Trpg Catalog" %></title>
+    <%= display_meta_tags(default_meta_tags) %>
```
karia: [blocking] **Every page renders `<title>卓の記録</title>`; `content_for :title` is ignored by meta-tags.**

The old layout read `content_for(:title)` into the `<title>` tag; `display_meta_tags` does not. meta-tags builds the title solely from the `:title` key in its own tag collection (`MetaTags::Renderer`), and nothing in the gem reads `content_for`. So the `<% content_for :title, ... %>` lines in `scenarios/index`, `scenarios/show`, and all the manage views have no effect.

Observed in the test env:
```
INDEX TITLE: "<title>卓の記録</title>"
SHOW TITLE:  "<title>卓の記録</title>"
```
Note `og:title` *is* correct on the detail page, so only the real `<title>` is wrong — which is the one Lighthouse's SEO audit and search result snippets use, and the plan's step 10 ("meta-tags でタイトルと description、OGP を出す") is not met.

Smallest fix: add `title: content_for(:title)` to `default_meta_tags` (the layout renders after the template, so it is populated by then), or switch the views to `set_meta_tags title: ...`.

app/views/layouts/application.html.erb
```diff
@@ -1,31 +1,41 @@
...
 
-    <%# Includes all stylesheet files in app/assets/stylesheets %>
     <%= stylesheet_link_tag :app, "data-turbo-track": "reload" %>
     <%= javascript_importmap_tags %>
+    <%= render "shared/analytics" %>
   </head>
 
-  <body>
-    <main class="container mx-auto mt-28 px-5 flex">
+  <body class="min-h-dvh flex flex-col">
```
karia, line 19: [should-fix] **`flash` is never rendered anywhere, so every `notice:` in the manage controllers is invisible.**

`grep -rn flash app/views app/controllers` finds only the `notice:` arguments in `Manage::ScenariosController` and `Manage::MastersController` — no template renders `flash`. For the masters screens this matters most: create/update/destroy redirect back to the same index with no other visible change, so the operator gets no confirmation that anything happened.

A `<%= flash.each ... %>` block here in the layout (or a `shared/_flash` partial) covers both areas at once.

app/views/manage/masters/_index.html.erb
```diff
@@ -0,0 +1,24 @@
+<% content_for :title, title %>
+<h1 class="font-display text-2xl"><%= title %></h1>
+
+<%= form_with model: record, url: index_path, class: "mt-6 flex gap-2" do |form| %>
```
karia: [blocking] **Editing a game system or author is broken: the form PATCHes to the collection path and 404s.**

`MastersController#edit` sets `@record` to a persisted record and re-renders this template. `form_with model: record, url: index_path` still derives the verb from the record (`html_options[:method] ||= :patch if model.persisted?`), so the rendered form is:

```html
<form action="/manage/authors" method="post"><input type="hidden" name="_method" value="patch" />
```

and `resources :authors, except: [:show]` has no PATCH on the collection. Reproduced against the running app:

```
EDIT STATUS: 200
PATCH COLLECTION STATUS: 404
```

Pass the member path when the record is persisted, e.g. `url: record.persisted? ? record_path.call(record) : index_path`. Nothing in the suite covers these two controllers (see the comment on `spec/requests/manage/scenarios_spec.rb`), which is why this got through.

app/views/shared/_analytics.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if show_analytics? %>
+  <script async src="https://www.googletagmanager.com/gtag/js?id=<%= ENV["GA_MEASUREMENT_ID"] %>"></script>
+  <script nonce="<%= content_security_policy_nonce %>">
```
karia, line 3: [blocking] **The CSP nonce is always empty, so this inline snippet is blocked and GA never initialises.**

`content_security_policy_nonce_generator` is `request.session.id.to_s`, and with the cookie session store `session.id` is `nil` unless something forces a session id. Nothing in Phase 1 does (there is no auth yet, and the CSRF token in the cookie store does not allocate an id). Actual response from the test env with `GA_MEASUREMENT_ID` set:

```
script-src 'self' https://www.googletagmanager.com ... 'nonce-'
GA TAG:   "<script nonce=\"\">"
CSP META: "<meta name=\"csp-nonce\" content=\"\" />"
```

`'nonce-'` is not a valid source expression and `nonce=""` matches nothing, so the browser refuses the `gtag('config', ...)` block; the external `gtag/js` loads (allowed by host) but is never configured. The `<script type="application/ld+json">` on the detail page has no nonce at all and is inline too, so it is refused by the same directive.

Switch the generator to a per-request value (`->(request) { SecureRandom.base64(16) }`) and add `nonce: true` to the JSON-LD `tag.script`, or move the gtag config into an asset-served file so no nonce is needed.

config/initializers/content_security_policy.rb
```diff
@@ -1,20 +1,34 @@
...
+    policy.connect_src :self, *GOOGLE_MEASUREMENT_ORIGINS, *GOOGLE_TAG_ORIGINS
+    policy.frame_src   :self, *GOOGLE_TAG_ORIGINS
     policy.frame_ancestors :none
     policy.base_uri    :self
     policy.form_action :self
   end
 
   config.content_security_policy_nonce_generator = ->(request) { request.session.id.to_s }
-  config.content_security_policy_nonce_directives = %w[ script-src style-src ]
+  config.content_security_policy_nonce_directives = %w[ script-src ]
```
karia: [should-fix] **Dropping `style-src` from the nonce directives breaks Turbo's progress-bar stylesheet.**

`policy.style_src :self` has no `unsafe-inline`, and Turbo injects its progress bar as an inline `<style>` element carrying the value of the `csp-nonce` meta tag (`grep -o csp-nonce turbo.min.js` hits in the pinned `turbo.min.js`). With `nonce_directives = %w[script-src]` that nonce is no longer honoured for styles, so the element is refused and the navigation progress bar silently never appears.

Either keep `style-src` in the list, or drop it deliberately and note why. (It is moot until the nonce itself is non-empty — see the comment on `shared/_analytics.html.erb`.)

app/helpers/structured_data_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+      "name" => scenario.title,
+      "genre" => "TRPG シナリオ",
+      "url" => scenario_url(scenario),
+      "description" => scenario_description(scenario),
+      "inLanguage" => "ja"
+    }
+    data["author"] = scenario.authors.map { |a| { "@type" => "Person", "name" => a.name } } if scenario.authors.any?
+    data["timeRequired"] = "PT#{scenario.duration_min_minutes}M" if scenario.duration_min_minutes.present?
+
+    tag.script(raw(JSON.pretty_generate(data)), type: "application/ld+json")
```
karia: [should-fix] **JSON-LD is injected raw, so a title containing `</script>` breaks out of the script element.**

`JSON.pretty_generate` does not apply Rails' HTML-entity escaping (that only happens in `#to_json`/`ActiveSupport::JSON.encode` via `escape_html_entities_in_json`), and `raw` disables the view's escaping. Reproduced with a scenario titled `X</script><script>alert(1)</script>`:

```
LD: "<script type=\"application/ld+json\">{\n  \"@context\": ... \"name\": \"X</script>"
RAW BREAKOUT PRESENT: true
```

The injected element ends up in the document. Today only an authenticated editor can set a title and `script-src` would refuse the injected inline script, so this is not remotely exploitable — but it is a stored-HTML-injection primitive one CSP change away from mattering, and it corrupts the structured data for any title that merely contains `<` or `&`.

Use `data.to_json` (Rails escapes `<`, `>`, `&` to `\u003c` etc.) or wrap with `json_escape`.

app/helpers/scenarios_helper.rb
```diff
@@ -0,0 +1,41 @@
+module ScenariosHelper
+  def player_count_label(scenario)
+    range = [ scenario.player_count_min, scenario.player_count_max ].compact
```
karia: [should-fix] **One-sided player counts are rendered as if they were exact, and a blank row claims 「制限なし」.**

`[min, max].compact` throws away which end was missing, so a scenario with `min: 4, max: nil` ("4 people or more") and one with `min: nil, max: 5` ("up to 5 people") both render as a flat count. Measured:

```
PLAYERS [4, 5, nil]      => "4〜5人"      # correct
PLAYERS [4, nil, nil]    => "4人"        # means 4+
PLAYERS [nil, 5, nil]    => "5人"        # means up to 5
PLAYERS [nil, nil, nil]  => "制限なし"
```

The schema makes the two columns independently nullable, so one-sided rows are representable and the editing form lets you enter them. Branch on which end is present (`"4人〜"` / `"〜5人"`).

Separately, the last row above is an assertion the data does not support: an unfilled player count and an explicit 「制限なし」 are indistinguishable in the output, while `duration_label` correctly says 「未設定」 for the same situation. Either keep the two consistent, or require `player_count_note` to be the thing that produces 「制限なし」.

karia (commented):
Third-party review of Phase 1. The shape matches the plan and the ADR (public read path, Pundit scope, seams for Phase 3/4, basic auth failing closed when the env vars are unset), and I could not find any route that leaks `preparation_note` — the meta description, JSON-LD and sitemap all only use public fields. I did find several things I could reproduce by running the app in the test environment: every page renders `<title>卓の記録</title>` because meta-tags ignores `content_for(:title)`; the CSP nonce is always empty so the GA snippet is dead on arrival; the master (system/author) edit form PATCHes to the collection path and 404s; JSON-LD is emitted unescaped and a title containing `</script>` breaks out of the script element; and flash notices are never rendered. Details inline, each with the observed output.

app/helpers/scenarios_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+    [ numeric, scenario.player_count_note ].compact_blank.join(" ").presence || "制限なし"
+  end
+
+  def duration_label(scenario)
+    range = [ scenario.duration_min_minutes, scenario.duration_max_minutes ].compact
+    return "未設定" if range.empty?
+
+    range.uniq.map { |minutes| humanized_minutes(minutes) }.join("〜")
```
karia: [should-fix] **`duration_label` mixes units inside one range and has the same one-sided problem.**

Each end is humanised independently, so the spreadsheet's 「30分～60分」 comes out as a range whose two ends use different units, and the comment above `humanized_minutes` claiming 「6〜7時間」 reads well is not what is produced. Measured:

```
DURATION [30, 60]    => "30分〜1時間"
DURATION [360, 420]  => "6時間〜7時間"
DURATION [660, 1080] => "11時間〜18時間"   # 「11～18時間」 OK
DURATION [120, nil]  => "2時間"           # means 2h+
DURATION [75, 100]   => "1.3時間〜1.7時間"
```

Three things worth deciding: pick the unit from the pair rather than per value (so 30–60 stays in minutes, or becomes 「0.5〜1時間」); mark one-sided ranges as open (`"2時間〜"`); and note that `hours.round(1)` on non-multiples of 6 minutes yields values like 「1.3時間」 that read worse than 「1時間15分」.

None of this is covered by a spec — there is no `spec/helpers` at all, and the system spec only asserts the single `[120, nil] => "2時間"` case.

app/controllers/manage/base_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    skip_after_action :verify_authorized
+
+    private
+      def authenticate_editor
+        username = ENV["MANAGE_USERNAME"]
+        password = ENV["MANAGE_PASSWORD"]
+        return head :unauthorized if username.blank? || password.blank?
+
+        authenticate_or_request_with_http_basic("Manage") do |given_user, given_password|
+          ActiveSupport::SecurityUtils.secure_compare(given_user, username) &
```
karia: [should-fix] **A malformed `Authorization` header raises a 500 instead of returning 401.**

`ActionController::HttpAuthentication::Basic.user_name_and_password` splits the decoded credentials on `":"`, and `authenticate` splats the result into this block. If the client sends base64 of a string with no colon, the block gets `given_password = nil` and `secure_compare` blows up. Reproduced with `Basic #{Base64.strict_encode64("editor")}`:

```
MALFORMED RAISED: NoMethodError: undefined method 'bytesize' for nil
```

An unauthenticated request should never be able to reach the error handler. `given_user.to_s` / `given_password.to_s` fixes it.

The rest of this method reads correctly to me: `secure_compare` (the digest-based one) is length-safe, `&` avoids the short-circuit timing leak, and returning 401 when the env vars are blank fails closed. Worth adding a spec for that blank-env branch — the current suite always sets both vars.

app/controllers/sitemaps_controller.rb
```diff
@@ -0,0 +1,8 @@
+class SitemapsController < ActionController::Base
+  def show
+    path = Rails.root.join("tmp/sitemaps/sitemap.xml.gz")
+    return head :not_found unless File.exist?(path)
+
+    send_file path, type: "application/xml", disposition: "inline", content_type: "application/gzip"
```
karia: [should-fix] **`content_type:` silently overrides `type:`, so `/sitemap.xml` serves gzip bytes labelled `application/gzip`.**

`ActionController::DataStreaming#send_file` applies `options[:type]` first and then `self.content_type = options[:content_type] if options.key?(:content_type)` (actionpack `data_streaming.rb:84`). The `type: "application/xml"` here is dead — the response ends up `application/gzip` with no `Content-Encoding: gzip`, so anything following the `Sitemap:` line in `robots.txt` gets a binary download at a `.xml` URL rather than XML.

Either send the compressed bytes with `type: "application/xml"` plus `response.headers["Content-Encoding"] = "gzip"`, or set `SitemapGenerator::Sitemap.compress = false` and serve the plain file. Also worth a request spec: the plan lists 「`/sitemap.xml` が生成され、公開シナリオの URL を列挙している」 as a verification step and nothing in the suite touches this route, including the 404-when-missing branch (which is what a fresh pod will hit until the rake task has run — the file lives in the container's `tmp/`).

app/views/scenarios/_preparation_note.html.erb
```diff
@@ -0,0 +1,9 @@
+<%# Phase 4 がここにネタバレ防止ボタンと準備情報を埋める。show.html.erb には触らない。 %>
+<% if scenario.preparation_note.present? %>
```
karia: [should-fix] **The seam holds, but the visibility decision is hardcoded here instead of going through `ScenarioPolicy#show_preparation_note?`.**

The partial does what the plan asks — it renders a placeholder, `show.html.erb` renders both partials with a `scenario:` local, and `_play_session_history` is genuinely empty, so Phase 3 and Phase 4 can each fill one without touching `show.html.erb`. But `ScenarioPolicy#show_preparation_note?` (added in this PR, returning `false`) is never called by anything, so it is dead code and the ADR's 「可視性の判定を 1 箇所に書く」 rule is not actually wired up: Phase 4 has to remember to introduce the call.

Writing it as `<% if scenario.preparation_note.present? && policy(scenario).show_preparation_note? %>` now, with the placeholder in the `else`, makes the seam self-enforcing and gives you a spec that pins the policy rather than the template.

For what it's worth I checked the other exits for this field and found none: `scenario_description` only uses systems/players/duration/synopsis, the JSON-LD carries the same description, the sitemap emits paths only, and there is no JSON view.

app/views/manage/scenarios/_link_row.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="grid gap-2 sm:grid-cols-[10rem_1fr_auto]">
+  <%= link.text_field :label, placeholder: "ラベル", class: "border border-rule bg-paper px-3 py-2 text-sm" %>
```
karia, line 3: [should-fix] **These inputs have no accessible name — placeholder text is not a label.**

`text_field :label` and `url_field :url` here rely entirely on `placeholder`, so a screen reader announces "edit text, blank" and the hint disappears as soon as the user types. Same pattern in `manage/masters/_index.html.erb:5` (`text_field :name, placeholder: "名前"`). Everywhere else in this PR the pairing is done properly (`_form.html.erb` labels every field), so this reads as an oversight rather than a choice.

`link.label :label, "ラベル", class: "sr-only"` (and likewise for `:url`) keeps the compact grid while giving each control a name.

spec/requests/manage/scenarios_spec.rb
```diff
@@ -0,0 +1,88 @@
+require "rails_helper"
+
+RSpec.describe "Manage::Scenarios" do
+  let(:credentials) { ActionController::HttpAuthentication::Basic.encode_credentials("editor", "secret") }
+
```
karia, line 5: [should-fix] **The manage area's shared controller has no coverage, which is exactly where the bug is.**

This file exercises `Manage::ScenariosController` well (401 without credentials, the whole nested create in one submission, the unprocessable re-render). But `Manage::MastersController` — the class two controllers share — has no spec at all, and its edit path 404s (see `manage/masters/_index.html.erb`). Two request specs against `/manage/authors` (edit renders, update persists) would have caught it.

Two other gaps in the same area:

- The `rack_test` driver in the system spec is the right call for what that spec claims (list → detail is plain links, and it does assert the rendered `2時間`/`★★★★★`/absence of the preparation note, so it would fail if the helpers or the partial regressed). But it means `nested_form_controller.js` — the only JavaScript in the PR, and the mechanism the whole 「行を足す」 flow depends on — is never executed by anything. Worth either a `:js` spec or an explicit note that it is manually verified.
- `spec/helpers` does not exist. `player_count_label` and `duration_label` encode most of the spreadsheet's edge cases (「制限なし」「1人推奨」「4～5人」「30分～60分」「11～18時間」and blank) and are currently pinned by exactly one assertion in the system spec.

app/views/layouts/application.html.erb
karia: Confirmed and fixed in 51c1785. Reproduced first: both `/` and `/scenarios/1` served `<title>卓の記録</title>`.

Fixed in the layout rather than by rewriting every view, since `content_for :title` is what the manage screens use too: `display_meta_tags(default_meta_tags.merge(title: content_for(:title).presence))`.

Now serves `シナリオ一覧 | 卓の記録` and `カタシロ | 卓の記録`, verified against the running app. A request spec asserts the exact title on the detail page.

app/views/layouts/application.html.erb
karia: Fixed in 51c1785. The layout now renders `flash` above `<main>` with `role="status"`, and `alert` is coloured with the seal accent.

app/views/manage/masters/_index.html.erb
karia: Confirmed and fixed in 51c1785. The form now posts to `record_path.call(record)` when persisted and to the collection otherwise, so PATCH lands on the member route.

You were right that the absence of specs is the real finding here. Added `spec/requests/manage/masters_spec.rb` with a shared example run against both GameSystem and Author: 401 without credentials, index, create, blank-name re-render, edit, **update through the member path**, destroy. 14 examples. The update example fails against the old form.

app/views/shared/_analytics.html.erb
karia: Confirmed and fixed in 51c1785. Reproduced: the header carried `'nonce-'` and every tag rendered `nonce=""`.

The generator is now `->(_request) { SecureRandom.base64(16) }`. Rails memoises the value per request, so the header and the tags agree. The JSON-LD block now carries the nonce too.

Verified against the running app: header has `'nonce-5xLQ5VlBJd2yfOjb1sHX3Q=='` and the tags carry the same value. A request spec extracts the nonce from the header and asserts the body contains it, so an empty nonce fails the suite.

config/initializers/content_security_policy.rb
karia: Fixed in 310e289. Restored `style-src` to `content_security_policy_nonce_directives` with a comment naming Turbo's progress bar as the reason, so it does not get dropped again by someone tidying the list.

app/helpers/structured_data_helper.rb
karia: Fixed in 51c1785. Now `JSON.generate(data).gsub("<", "\\u003c")` before marking it safe, which neutralises `</script` without changing what a JSON-LD parser reads.

A request spec sets the title to `危険</script><script>alert(1)</script>` and asserts the escaped sequence is absent from the body.

app/helpers/scenarios_helper.rb
karia: Fixed in 51c1785. Extracted a `bounded_label` helper that distinguishes the four states: both ends equal → `3人`; a real range → `4人〜5人`; minimum only → `4人`; maximum only → `5人まで`.

Also made the empty case consistent with `duration_label`: both ends blank and no note now says `未設定` rather than asserting `制限なし`. The scenarios that really are unlimited carry that in `player_count_note`, which still takes precedence.

app/helpers/scenarios_helper.rb
karia: Fixed in 51c1785. `duration_label` now picks one unit for the whole range instead of formatting each end independently: minutes when the larger end is under two hours, hours otherwise.

`30〜60` → `30分〜60分`, matching the spreadsheet. `75〜100` → `75分〜100分` rather than `1.3時間〜1.7時間`. `360〜480` → `6時間〜8時間`. `660〜1080` → `11時間〜18時間`.

Added `spec/helpers/scenarios_helper_spec.rb`, 17 examples, driven by the actual spreadsheet values you listed.

app/controllers/manage/base_controller.rb
karia: Fixed in 51c1785 with `.to_s` on both sides before `secure_compare`.

Added a spec that sends `Basic base64("editor")` with no colon and asserts 401 rather than 500.

app/controllers/sitemaps_controller.rb
karia: Fixed in 51c1785, by removing the compression rather than adding `Content-Encoding`. `SitemapGenerator::Sitemap.compress = false` and the controller sends `tmp/sitemaps/sitemap.xml` with `type: "application/xml"` and no override.

Verified: `/sitemap.xml` returns 200 with `content_type=application/xml`. The site is small enough that the gzip saved nothing worth the failure mode.

app/views/scenarios/_preparation_note.html.erb
karia: Fixed in 51c1785. The partial now reads `scenario.preparation_note.present? && !policy(scenario).show_preparation_note?`, so `ScenarioPolicy#show_preparation_note?` is the single point that decides. Phase 4 changes the policy and the template follows.

app/views/manage/scenarios/_link_row.html.erb
karia: Fixed in 51c1785. Added `sr-only` labels for both fields in `_link_row`, and gave the masters form a visible label instead of a bare placeholder.

spec/requests/manage/scenarios_spec.rb
karia: Fixed in 51c1785 for the two gaps that mattered: `spec/requests/manage/masters_spec.rb` (14 examples, shared between GameSystem and Author) and `spec/helpers/scenarios_helper_spec.rb` (17 examples).

`nested_form_controller.js` stays uncovered. Exercising it needs a JS-capable driver and a browser in CI, which Phase 0 deliberately kept out. Recording it as a known gap rather than pretending otherwise.

Suite is now 87 examples, up from 52.

PR #11, opened 2026-08-10
Phase 2: Google sign-in, people, roles and groups

karia:
## What

Implements [Phase 2](docs/plans/2026-08-09-phase2-authentication.md).

- Google sign-in via OmniAuth. The request phase is POST-only (`omniauth-rails_csrf_protection`); the callback accepts the GET that Google sends back
- `Person`, `PersonRole`, `Group`, `GroupMembership`, `User`. Roles live on `Person`, so swapping the Google account later does not drop them
- `pundit_user` is `current_person`, not `current_user`
- The `/manage` area's temporary basic auth is gone, replaced by role checks. Admins get everything; GMs get scenarios and the master tables but not people or groups
- Admin screens for people, groups, and linking a Google account to a person
- The analytics and ad gate now reads the real sign-in state

## Notable

- An account with no `Person` is a normal state, not an error. Every user passes through it: they sign in, see only the public area, and an admin links them afterwards. Specs cover that transition in both directions.
- Denied requests answer 404 rather than 403, so a hidden page does not confirm it exists.
- `spec/policies/authorization_matrix_spec.rb` fixes the whole grid in one place — anonymous, signed in but unlinked, linked with no role, GM, admin — because ADR-0001 treats an authorization gap as a disclosure.
- `bin/rails admin:grant EMAIL=... NAME=...` creates the first admin, which the admin screens cannot do.

## Verification

- [x] `bin/rspec` — 129 examples, 0 failures
- [x] `prek run --all-files`
- [x] GET on `/auth/google_oauth2` answers 404, so another site cannot start the flow
- [x] A provider failure redirects instead of raising, and creates no user

## Not in this PR

The Google OAuth client itself. `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` are read from the environment and the redirect URIs are documented in `.env.example`; registering the client in Google Cloud is a manual step before deploying.


app/controllers/manage/masters_controller.rb
```diff
@@ -13,12 +14,13 @@ def new
     end
```
karia: [blocking] **`#new` never calls `authorize`, so `verify_authorized` raises an exception that is not rescued — guaranteed 500 on three shipped routes.**

`new` only redirects; the `after_action :verify_authorized` still runs and raises `Pundit::AuthorizationNotPerformedError`. `ApplicationController` rescues only `Pundit::NotAuthorizedError`, and `AuthorizationNotPerformedError` is *not* a subclass of it, so nothing catches it.

Reproduced:
```
GET /manage/game_systems/new  as GM    -> Pundit::AuthorizationNotPerformedError
GET /manage/groups/new        as admin -> Pundit::AuthorizationNotPerformedError
```
`resources :game_systems/:authors/:groups, except: [ :show ]` all route `new`, so all three are reachable. No spec covers `new`, which is why CI is green.

Fix: `authorize model_class` before the redirect (or `skip_after_action :verify_authorized, only: :new`). Consider also `except: [ :show, :new ]` in routes, since the index already carries the create form.

app/controllers/manage/people_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  class PeopleController < BaseController
+    before_action :set_person, only: %i[edit update destroy]
+
+    def index
+      authorize Person
+      @people = policy_scope(Person).includes(:person_roles, :groups, :user)
+      @person = Person.new
+    end
+
+    def new
```
karia: [should-fix] **`new` authorizes and then renders a template that does not exist.**

`app/views/manage/people/` contains only `_form`, `edit` and `index`. Reproduced: `GET /manage/people/new` as an admin answers **406** (`ActionController::MissingExactTemplate`).

Either add `new.html.erb`, or drop the action and use `except: [ :show, :new ]` — the index page already renders the create form via `@person = Person.new`.

config/routes.rb
```diff
@@ -7,12 +7,20 @@
   # get "manifest" => "rails/pwa#manifest", as: :pwa_manifest
   # get "service-worker" => "rails/pwa#service_worker", as: :pwa_service_worker
 
+  # 開始は POST 限定（omniauth-rails_csrf_protection）。コールバックは Google からの GET。
+  match "/auth/:provider/callback", to: "sessions#create", via: [ :get, :post ]
```
karia: [should-fix] **The wildcard `:provider` lets an anonymous visitor trigger an unhandled `KeyError`.**

OmniAuth's middleware only intercepts `/auth/google_oauth2/callback`. Any other provider name falls straight through to `sessions#create`, where `request.env.fetch("omniauth.auth")` raises.

Reproduced: `GET /auth/bogus/callback` -> `KeyError: key not found: "omniauth.auth"` (500 + exception noise in the logs, from an unauthenticated request).

Fix: constrain the segment, e.g. `match "/auth/:provider/callback", to: "sessions#create", via: [ :get, :post ], constraints: { provider: /google_oauth2/ }`.

(For the record, the GET callback itself is safe: with the real strategy a callback carrying no `state` is rejected — I saw `csrf_detected: CallbackError` and a 302 to the failure path.)

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,27 @@
+class SessionsController < ApplicationController
+  skip_after_action :verify_authorized
+
+  def create
+    user = User.from_google(request.env.fetch("omniauth.auth"))
```
karia, line 5: [should-fix] **`from_google` can raise inside the callback, and there is no rescue — a provider hiccup becomes a 500 instead of the failure page.**

`User.from_google` calls `save!`. Reproduced:
```ruby
User.from_google(OmniAuth::AuthHash.new(provider: "google_oauth2", uid: nil, info: { email: "a@b.c" }))
# => ActiveRecord::RecordInvalid
```
A missing `info` is handled fine (`auth.info&.email` -> `email: nil`), but a blank uid, or a `RecordNotUnique` from a concurrent first sign-in of the same account, both escape as 500s.

Fix: `rescue ActiveRecord::ActiveRecordError` here and fall through to the same redirect `failure` uses, so the user sees "ログインできませんでした" rather than an error page.

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    redirect_to root_path, notice: "ログアウトしました"
+  end
+
+  def failure
+    redirect_to root_path, alert: "ログインできませんでした"
+  end
+
+  private
+    def after_sign_in_path(user)
+      return root_path if user.linked?
```
karia: [optional] **Both branches of `after_sign_in_path` return `root_path`.**

The `linked?` check has no effect. Either drop the method and redirect to `root_path` directly, or make the unlinked branch land somewhere that explains the state (the layout already shows "未登録").

lib/tasks/admin.rake
```diff
@@ -0,0 +1,15 @@
+# 最初の管理者は管理画面からは作れないため、ここで作る。
+namespace :admin do
+  desc "Link a Google account to a person and grant the admin role (EMAIL=, NAME=)"
+  task grant: :environment do
+    email = ENV.fetch("EMAIL")
+    name = ENV.fetch("NAME", email.split("@").first)
+
+    user = User.find_by(email:) or abort("No user has signed in with #{email} yet")
```
karia: [should-fix] **The admin-grant task looks the account up by `email`, which is neither unique nor an identity — and it grants `admin`.**

`users.email` has no uniqueness validation and no unique index. Reproduced:
```ruby
User.create!(google_uid: "u1", email: "same@example.com")
User.create!(google_uid: "u2", email: "same@example.com")  # both persist
```
That state is reachable in practice: a Workspace address that is deleted and re-issued produces a second account with a new `google_uid` while the stale row keeps the old address. `find_by` then picks an arbitrary row and grants admin to it.

The plan says to key on the uid — 「指定した `google_uid` の `User` を Person に紐づけて管理者権限を与える」 — and `User.from_google` correctly treats uid as the identity. This task is the one place that departs from that. Suggest `GOOGLE_UID=` (optionally falling back to email but aborting when more than one row matches).

app/policies/user_policy.rb
```diff
@@ -0,0 +1,10 @@
+class UserPolicy < ApplicationPolicy
+  def index? = admin?
```
karia, line 2: [should-fix] **The policy that binds a Google account to a Person — and therefore to its roles — has zero test coverage.**

Mutation test: I changed both `index?` and `update?` from `admin?` to `editor?` and ran the full suite — **129 examples, 0 failures**. Nothing in `spec/` ever exercises `manage_users_path` as a GM (grep shows only admin `PATCH manage_user_path` in `people_spec.rb`), and `UserPolicy` is absent from the authorization matrix.

This is the highest-leverage policy in the PR: with it open to GMs, a GM could point a Google account at an admin `Person` and inherit the admin role. The code is correct today; it is just unguarded against regression.

Add request specs: GM -> 404 on `GET /manage/users` and on `PATCH /manage/users/:id`.

spec/policies/authorization_matrix_spec.rb
```diff
@@ -0,0 +1,82 @@
+require "rails_helper"
+
+# ADR-0001 は認可の漏れをそのまま情報漏えいとして扱う。役割ごとの可否を一覧で固定する。
+RSpec.describe "Authorization matrix" do
+  let(:anonymous) { nil }
+  let(:unlinked) { nil }
```
karia, line 7: [should-fix] **The matrix claims five viewer states but exercises four; `unlinked` is defined and never used.**

`let(:unlinked) { nil }` never appears in any example, and at the policy layer it is indistinguishable from `anonymous` anyway — the unlinked/anonymous distinction lives in `current_person`, not in the policies. So the state the ADR cares most about (signed in, public area only) is only covered by `manage/scenarios_spec.rb`.

Also missing from the matrix: `UserPolicy` entirely, `AuthorPolicy`/`GameSystemPolicy` (both of which deliberately admit GMs), and `create?`/`destroy?` for `PersonPolicy`/`GroupPolicy`. A table-driven `[state] x [policy] x [action]` loop would make the exhaustiveness checkable rather than asserted.

karia (commented):
Third-party review of the Phase 2 auth work. The core model is sound: identity is keyed on `google_uid` (not email), `reset_session` runs before the session is populated, the request phase is POST-only with `omniauth-rails_csrf_protection` active, the OAuth callback rejects a missing `state` (verified: `csrf_detected`), `pundit_user` is `current_person`, `GroupsController` really does resolve to `GroupPolicy`, and `users.person_id` has a real unique index so two accounts cannot share a Person. Brakeman is clean and all 129 examples pass. What I did find: two reproducible 500s on routes this PR ships, an admin-grant task keyed on a non-identity attribute, two N+1s, and — most importantly — three mutations of security-relevant code that leave the entire suite green (`UserPolicy` opened to GMs, `reset_session` deleted, `on_failure` deleted). Details inline; every claim below was reproduced by running it.

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    it "signs the same account in again without creating another" do
+      sign_in
+      delete session_path
+
+      expect { sign_in }.not_to change(User, :count)
+    end
+
+    it "reports a failure from the provider instead of raising" do
```
karia: [should-fix] **This example passes with the failure handling removed.**

Mutation test: I deleted `OmniAuth.config.on_failure = ...` from `config/initializers/omniauth.rb` and ran this file — **5 examples, 0 failures**. Two reasons: OmniAuth's default failure redirect also lands on the `get "/auth/failure"` route, and `redirect_to(root_path).or have_http_status(:ok)` matches the final 200 on the root page regardless of what happened in between.

Fix: drop the `.or`, do a single `follow_redirect!`, and assert the intermediate hop (`expect(response).to redirect_to(root_path)`) plus `flash[:alert]`.

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  describe "signing in" do
+    it "refuses to start the flow over GET, so another site cannot trigger it" do
+      get "/auth/google_oauth2"
+
+      expect(response).to have_http_status(:not_found)
+    end
+
+    it "creates an account that is not linked to a person yet" do
```
karia, line 30: [should-fix] **Nothing tests session fixation, the one thing `reset_session` is there for.**

Mutation test: I removed `reset_session` from `SessionsController#create` and ran the whole suite — **129 examples, 0 failures**. The placement in the controller is correct (reset, then write `user_id`), but it is unguarded.

Add an example that captures the session cookie / `request.session.id` before `post "/auth/google_oauth2"` and asserts it changed afterwards.

spec/requests/sessions_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    OmniAuth.config.test_mode = false
+  end
+
+  def sign_in
+    post "/auth/google_oauth2"
+    follow_redirect!
+  end
+
+  describe "signing in" do
+    it "refuses to start the flow over GET, so another site cannot trigger it" do
```
karia, line 24: [optional] **This proves `allowed_request_methods`, not the CSRF protection the plan names.**

`config/environments/test.rb` sets `allow_forgery_protection = false`, so `OmniAuth::RailsCsrfProtection::TokenVerifier` is a no-op in specs — I confirmed the rendered login `button_to` carries no `authenticity_token` under test. The plan calls out omniauth-rails_csrf_protection as the mitigation, so it is worth one example that re-enables forgery protection and asserts a tokenless `POST /auth/google_oauth2` is rejected.

app/views/manage/users/index.html.erb
```diff
@@ -0,0 +1,24 @@
...
+<ul class="mt-6 divide-y divide-rule border border-rule bg-surface text-sm">
+  <% @users.each do |user| %>
+    <li class="flex flex-wrap items-center justify-between gap-3 p-3">
+      <div>
+        <p class="font-mono text-xs"><%= user.email.presence || "メールアドレス不明" %></p>
+        <p class="text-xs text-muted">初回ログイン <%= l user.created_at, format: :short %></p>
+      </div>
+      <%= form_with model: [ :manage, user ], class: "flex items-center gap-2" do |form| %>
+        <%= form.label :person_id, "紐づけ先", class: "sr-only" %>
+        <%= form.collection_select :person_id, Person.all, :id, :display_name,
```
karia: [should-fix] **`Person.all` runs inside the per-user loop — one people query per row.**

The controller carefully does `includes(:person)`, and then this rebuilds the whole Person collection for every `<li>`. Measured with 3 users: 9 statements total, **6 of them `FROM "people"`**.

Hoist it: set `@people = Person.all` (or `policy_scope(Person)`) in `Manage::UsersController#index` and pass that to `collection_select`.

app/controllers/manage/people_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      authorize @person
+    end
+
+    def create
+      @person = authorize Person.new(person_params)
+
+      if @person.save
+        redirect_to manage_people_path, notice: "#{@person.display_name} を登録しました"
+      else
+        @people = policy_scope(Person)
```
karia: [should-fix] **The create-failure path drops the `includes` that the index action added, reintroducing the N+1.**

`index` uses `policy_scope(Person).includes(:person_roles, :groups, :user)`; this line re-renders the same template with a bare `policy_scope(Person)`, and `index.html.erb` touches `item.roles`, `item.groups` and `item.user` for every row.

Measured with 4 people: happy path **9** statements, failure path **17** (5 x `person_roles`, 5 x `groups`, 5 x `users`). Extract the scope into a private method and use it from both places.

app/controllers/manage/masters_controller.rb
```diff
@@ -4,6 +4,7 @@ class MastersController < BaseController
     before_action :set_record, only: %i[edit update destroy]
 
     def index
+      authorize model_class
       @records = model_class.all
```
karia: [should-fix] **`index` uses `model_class.all` instead of `policy_scope`, and there is no `verify_policy_scoped` anywhere.**

`Manage::PeopleController#index` and `Manage::UsersController#index` both go through `policy_scope`; this one does not, which makes `GroupPolicy::Scope` dead code (verified: an admin's `GET /manage/groups` never touches it). No leak today, because `GroupPolicy::Scope` is all-or-nothing on `admin?` and `require_editor` + `authorize` already gate the action.

The risk is Phase 3: ADR-0001 puts the whole PlaySession visibility rule in a `Scope`, and without `after_action :verify_policy_scoped` on the index actions a resource that *does* need row filtering can silently skip it. Consider `@records = policy_scope(model_class)` here plus `verify_policy_scoped` on collection actions.

app/models/person.rb
```diff
@@ -0,0 +1,29 @@
...
+  has_one :user, dependent: :nullify
+  has_many :person_roles, dependent: :destroy
+  has_many :group_memberships, dependent: :destroy
+  has_many :groups, through: :group_memberships
+
+  validates :display_name, presence: true
+
+  default_scope { order(:display_name) }
+
+  scope :sharing_a_group_with, ->(person) {
```
karia, line 17: [should-fix] **`sharing_a_group_with` excludes the person themself, which is a trap for the Phase 3 visibility rule.**

The SQL is correct and indexable — I checked the generated plan shape:
```sql
SELECT people.* FROM people
WHERE people.id != $1
  AND people.id IN (SELECT group_memberships.person_id FROM group_memberships WHERE group_memberships.group_id = $2)
ORDER BY people.display_name ASC
```
the subquery hits `index_group_memberships_on_group_id`, and a person with no groups correctly yields zero rows.

Two things to settle before Phase 3 builds on it:
1. ADR-0001's rule is "参加者である **または** 参加者の誰かと同じグループ". Written as `PlaySession.where(participant_id: Person.sharing_a_group_with(viewer))` the `where.not(id: person.id)` silently drops the viewer's own sessions, so the OR-branch becomes load-bearing. A companion `sharing_a_group_with_or_is(person)` (or dropping the exclusion and letting the caller decide) removes the footgun.
2. `person.group_ids` executes eagerly, so the scope is two round trips and cannot be composed against a preloaded viewer. A self-join on `group_memberships` keeps it to one statement.

Minor: `sharing_a_group_with(nil)` raises `NoMethodError`, and for an unsaved Person it degenerates to `id IS NOT NULL AND ... WHERE 1=0` — harmless, but worth a guard clause given a policy will call it.

app/controllers/manage/people_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        render :index, status: :unprocessable_content
+      end
+    end
+
+    def update
```
karia, line 26: [should-fix] **Nothing stops the last admin from removing their own admin role and locking everyone out.**

Reproduced end to end: the sole admin `PATCH`es their own record with `person[roles][]=""`, gets a 302, and afterwards `PersonRole.where(name: "admin").count == 0`; their very next `GET /manage/people` answers 404. The same applies to `destroy` on their own Person. Recovery then requires shell access to `bin/rails admin:grant`.

Suggest a guard in `PersonPolicy` (or a model validation) refusing an update/destroy that would drop the last `admin` role.

On the mass-assignment question, for the record: `roles: []` and `group_ids: []` are not an escalation path for a GM — `PersonPolicy` is `admin?` on every action and `require_editor` cannot be reached around (I confirmed a GM gets 404 on `/manage/people` and on `/manage/users`, and that weakening `require_editor` to `current_person.present?` is caught by a spec). I also confirmed `update` is safe against a partial write: Rails wraps `assign_attributes` + `save` in one transaction, so a failed `display_name` validation rolls back the `person_roles` and `group_memberships` replacement (verified for both `Person#roles=` and `Group#person_ids=`).

app/controllers/manage/users_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  class UsersController < BaseController
+    def index
+      authorize User
+      @users = policy_scope(User).includes(:person).order(:created_at)
+    end
+
+    def update
+      user = authorize User.find(params[:id])
+
+      if user.update(user_params)
```
karia, line 12: [optional] **The two-accounts-one-person race is closed at the database, but surfaces as a 500 rather than the 422 this branch produces.**

`users.person_id` has a genuine unique index and the column is nullable, so Postgres permits many unlinked accounts and at most one link per Person — the model validation and the DB agree, and the existing "refuses to link one person to two accounts" spec passes for the right reason. Two concurrent `PATCH`es, however, both pass validation and the loser raises `ActiveRecord::RecordNotUnique`.

Low likelihood (admin-only screen), but `rescue ActiveRecord::RecordNotUnique` here and rendering the same 422 would make the two paths behave identically.

app/controllers/concerns/authentication.rb
```diff
@@ -2,11 +2,26 @@ module Authentication
...
+
+  # Person に紐づいていないユーザーは nil を返す。ログイン必須エリアはこれで閉じる。
+  def current_person
+    current_user&.person
   end
+
+  def signed_in? = current_user.present?
+
+  private
+    def require_person
```
karia: [optional] **`require_person` is defined and never used.**

No controller references it — grep finds only this definition. The plan lists it as part of this concern (「`current_user` と `current_person`、ログイン必須の before_action」), and the verification item 「Person 未紐づけのユーザーが、ログイン必須のパスにアクセスすると弾かれる」 is currently satisfied only incidentally, by the manage area answering 404 rather than by this redirect-with-alert.

That is defensible while the only gated area is `/manage`, but either wire it up or note in the plan that it lands with Phase 3, so it does not read as an active guard.

app/views/manage/people/_form.html.erb
```diff
@@ -0,0 +1,41 @@
+<%= form_with model: [ :manage, person ], class: "mt-6 space-y-4 border border-rule bg-surface p-5" do |form| %>
+  <% if person.errors.any? %>
+    <p class="text-sm text-seal"><%= person.errors.full_messages.join("、") %></p>
```
karia, line 3: [optional] **Validation errors render as `Translation missing` text.**

`config.i18n.available_locales = [ :ja ]` with no rails-i18n and no `errors.messages` in `config/locales/ja.yml`. Reproduced — this is the literal body of the 422 response after posting a blank display name:

```
<p class="text-sm text-seal">Display name Translation missing. Options considered were:
- ja.activerecord.errors.models.person.attributes.display_name.blank
- ja.activerecord.errors.models.person.blank
- ja.activerecord.errors.messages.blank
...</p>
```

Pre-existing (the masters form has the same problem), but this PR adds the forms an admin will actually hit and already extends `ja.yml`. Adding `rails-i18n`, or `ja.errors.messages` plus `activerecord.attributes`, would close it in one place.

PR #15, opened 2026-08-10
Phase 3: session list and detail, gated by shared groups

karia:
## What

Implements [Phase 3](docs/plans/2026-08-09-phase3-session-area.md).

- `PlaySession` and `Participation`. Date and time are separate columns, because a planned session often has a date but no fixed start. `status` is stored rather than derived, because a past-dated session can be cancelled.
- Session list and detail, visible only to people who share a group with a participant
- The session history on a scenario page, filling the empty partial Phase 1 left. `scenarios/show.html.erb` is untouched.
- A GM-facing editing screen with a row per participant carrying role, character name and character sheet URL

## Visibility

The rule lives only in `PlaySessionPolicy::Scope`: the viewer's Person is a participant, or shares a group with one. Admins see everything. `show` looks the record up *through* the scope rather than re-testing the condition, so there is one place to get wrong.

Written as `EXISTS` over participations rather than a join, so a session with several qualifying participants is still returned once — there is a spec for that.

`spec/policies/play_session_policy_spec.rb` fixes all six viewer states the plan names, plus two leak cases.

## Verification

- [x] `bin/rspec` — 199 examples, 0 failures
- [x] `prek run --all-files`
- [x] Query counts measured: index 9, detail 8, scenario page 15, every query distinct — no N+1
- [x] A session detail answers 404 outside the scope rather than 403, so it does not confirm the record exists

## Notes

The manage screen deliberately lists every session regardless of the viewing GM's groups; it is for maintenance, and that is a different judgement from the reader-facing scope. Stated in the UI.

### DIFF
--- a/app/controllers/manage/play_sessions_controller.rb
+++ b/app/controllers/manage/play_sessions_controller.rb
@@ -1,0 +1,56 @@
+module Manage
+  class PlaySessionsController < BaseController
+    before_action :set_play_session, only: %i[edit update destroy]
+
+    def index
+      authorize PlaySession
+      # 編集者は保守のためすべての回を見る。閲覧側の Scope とは別の判断。
+      @play_sessions = PlaySession.includes(:scenario, participations: :person).newest_first
+    end
+
+    def edit
+      authorize @play_session
+    end
+
+    def create
+      @play_session = authorize PlaySession.new(play_session_params)
+
+      if @play_session.save
+        redirect_to manage_play_sessions_path, notice: "セッションを登録しました"
+      else
+        @play_sessions = PlaySession.includes(:scenario).newest_first
+        render :index, status: :unprocessable_content
+      end
+    end
+
+    def update
+      authorize @play_session
+
+      if @play_session.update(play_session_params)
+        redirect_to manage_play_sessions_path, notice: "セッションを更新しました"
+      else
+        render :edit, status: :unprocessable_content
+      end
+    end
+
+    def destroy
+      authorize @play_session
+      @play_session.destroy!
+      redirect_to manage_play_sessions_path, notice: "セッションを削除しました"
+    end
+
+    private
+      def set_play_session
+        @play_session = PlaySession.find(params[:id])
+      end
+
+      def play_session_params
+        params.expect(
+          play_session: [
+            :scenario_id, :played_on, :started_at, :status, :recording_url, :note,
+            { participations_attributes: [ [ :id, :person_id, :role, :character_name, :character_sheet_url, :position, :_destroy ] ] }
+          ]
+        )
+      end
+  end
+end
--- a/app/controllers/manage/play_sessions_controller.rb
+++ b/app/controllers/manage/play_sessions_controller.rb
@@ -3,9 +3,9 @@ module Manage
     before_action :set_play_session, only: %i[edit update destroy]
 
     def index
-      authorize PlaySession
+      authorize PlaySession, :manage?
       # 編集者は保守のためすべての回を見る。閲覧側の Scope とは別の判断。
-      @play_sessions = PlaySession.includes(:scenario, participations: :person).newest_first
+      @play_sessions = maintained_sessions
     end
 
     def edit
@@ -18,7 +18,7 @@ module Manage
       if @play_session.save
         redirect_to manage_play_sessions_path, notice: "セッションを登録しました"
       else
-        @play_sessions = PlaySession.includes(:scenario).newest_first
+        @play_sessions = maintained_sessions
         render :index, status: :unprocessable_content
       end
     end
@@ -40,6 +40,11 @@ module Manage
     end
 
     private
+      # 保守用なので公開側の Scope は通さない。入口は require_editor と manage? で守る。
+      def maintained_sessions
+        PlaySession.includes(:scenario, participations: :person).newest_first
+      end
+
       def set_play_session
         @play_session = PlaySession.find(params[:id])
       end
--- a/app/helpers/play_sessions_helper.rb
+++ b/app/helpers/play_sessions_helper.rb
@@ -1,9 +1,14 @@
 module PlaySessionsHelper
+  # 行ごとに Person.all を引くと参加者の数だけクエリが増える。
+  def people_for_select
+    @people_for_select ||= Person.all.to_a
+  end
+
   def play_session_schedule(session)
     return "日程未定" if session.played_on.blank?
 
     date = l(session.played_on, format: :long)
-    session.started_at.present? ? "#{date} #{l(session.started_at, format: :short)}" : date
+    session.started_at.present? ? "#{date} #{session.started_at.strftime("%H:%M")}" : date
   end
 
   def play_session_status_label(session)
--- a/app/helpers/play_sessions_helper.rb
+++ b/app/helpers/play_sessions_helper.rb
@@ -1,0 +1,16 @@
+module PlaySessionsHelper
+  def play_session_schedule(session)
+    return "日程未定" if session.played_on.blank?
+
+    date = l(session.played_on, format: :long)
+    session.started_at.present? ? "#{date} #{l(session.started_at, format: :short)}" : date
+  end
+
+  def play_session_status_label(session)
+    t("play_sessions.statuses.#{session.status}")
+  end
+
+  def participation_role_label(participation)
+    t("play_sessions.roles.#{participation.role}")
+  end
+end
--- a/app/models/participation.rb
+++ b/app/models/participation.rb
@@ -1,0 +1,9 @@
+class Participation < ApplicationRecord
+  enum :role, { gm: 0, player: 1, sub_keeper: 2 }, validate: true
+
+  belongs_to :play_session
+  belongs_to :person
+
+  validates :person_id, uniqueness: { scope: :play_session_id }
+  validates :character_sheet_url, http_url: true
+end
--- a/app/models/play_session.rb
+++ b/app/models/play_session.rb
@@ -5,12 +5,24 @@ class PlaySession < ApplicationRecord
   has_many :participations, -> { order(:position, :id) }, dependent: :destroy, inverse_of: :play_session
   has_many :people, through: :participations
 
-  accepts_nested_attributes_for :participations, allow_destroy: true, reject_if: :all_blank
+  # person_id が空の行は増やしただけの行なので捨てる。role には既定値があり all_blank では消えない。
+  accepts_nested_attributes_for :participations, allow_destroy: true,
+    reject_if: ->(attrs) { attrs["person_id"].blank? }
 
   validates :recording_url, http_url: true
+  validate :participants_are_distinct
 
   # 日付が無い回を末尾に固定する。データベース既定の NULL の並びに任せない。
-  scope :newest_first, -> { order(Arel.sql("played_on DESC NULLS LAST"), started_at: :desc, id: :desc) }
+  scope :newest_first, -> {
+    order(Arel.sql("played_on DESC NULLS LAST"), Arel.sql("started_at DESC NULLS LAST"), id: :desc)
+  }
 
-  def gm = participations.find { |p| p.gm? }&.person
+  private
+    # 一意制約は DB にもあるが、同じ送信内に重複があると例外になる前に弾く必要がある。
+    def participants_are_distinct
+      ids = participations.reject(&:marked_for_destruction?).filter_map(&:person_id)
+      return if ids.uniq.size == ids.size
+
+      errors.add(:base, "同じ人を複数の行に指定できません")
+    end
 end
--- a/app/models/play_session.rb
+++ b/app/models/play_session.rb
@@ -1,0 +1,16 @@
+class PlaySession < ApplicationRecord
+  enum :status, { scheduled: 0, played: 1, cancelled: 2 }, validate: true
+
+  belongs_to :scenario
+  has_many :participations, -> { order(:position, :id) }, dependent: :destroy, inverse_of: :play_session
+  has_many :people, through: :participations
+
+  accepts_nested_attributes_for :participations, allow_destroy: true, reject_if: :all_blank
+
+  validates :recording_url, http_url: true
+
+  # 日付が無い回を末尾に固定する。データベース既定の NULL の並びに任せない。
+  scope :newest_first, -> { order(Arel.sql("played_on DESC NULLS LAST"), started_at: :desc, id: :desc) }
+
+  def gm = participations.find { |p| p.gm? }&.person
+end
--- a/app/policies/play_session_policy.rb
+++ b/app/policies/play_session_policy.rb
@@ -3,6 +3,9 @@ class PlaySessionPolicy < ApplicationPolicy
   def show? = Scope.new(person, PlaySession).resolve.exists?(record.id)
 
   # シナリオと同じく、編集できるのは管理者と GM。
+  # 管理画面は公開側の index? とは別の判断。編集者だけが入る。
+  def manage? = editor?
+
   def create? = editor?
   def update? = editor?
   def destroy? = editor?
@@ -14,18 +17,17 @@ class PlaySessionPolicy < ApplicationPolicy
 
       # 本人が参加している回と、参加者の誰かと同じグループに属する回。
       # 行が膨らまないよう結合ではなく EXISTS で書く。
-      scope.where(
-        Participation.where("participations.play_session_id = play_sessions.id")
-          .where(person_id: visible_person_ids).arel.exists
-      )
+      mine = Participation.where("participations.play_session_id = play_sessions.id")
+
+      scope.where(mine.where(person_id: peer_person_ids).arel.exists)
+        .or(scope.where(mine.where(person_id: person.id).arel.exists))
     end
 
     private
-      def visible_person_ids
-        peers = GroupMembership.where(group_id: GroupMembership.where(person_id: person.id).select(:group_id))
+      # 同じグループの人の id。本人は所属が無くても参加者として見えるので別に足す。
+      def peer_person_ids
+        GroupMembership.where(group_id: GroupMembership.where(person_id: person.id).select(:group_id))
           .select(:person_id)
-
-        Person.where(id: peers).or(Person.where(id: person.id)).select(:id)
       end
   end
 end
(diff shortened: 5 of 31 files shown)


app/views/manage/play_sessions/edit.html.erb
```diff
@@ -0,0 +1,4 @@
+<% content_for :title, "セッションの編集" %>
+<h1 class="font-display text-2xl"><%= @play_session.scenario.title %> のセッション</h1>
```
karia: [blocking] **A failed update 500s instead of re-rendering the form.**

`update` renders `:edit` on validation failure, and this line dereferences `@play_session.scenario.title` unguarded. The scenario select in `_form` carries `include_blank: "選択してください"`, so a GM can clear it and submit — `belongs_to :scenario` then fails validation, `render :edit` runs, and `scenario` is nil.

Reproduced:

```
patch manage_play_session_path(s), params: { play_session: { scenario_id: "" } }
=> ActionView::Template::Error: undefined method 'title' for nil
```

No spec covers the update-failure path — `"updates a session"` only exercises the happy path, and the create-failure spec renders `:index`, not `:edit`. Suggest `@play_session.scenario&.title` (or a title that does not depend on the association) plus a request spec asserting 422 on a blanked `scenario_id`.

app/models/participation.rb
```diff
@@ -0,0 +1,9 @@
+class Participation < ApplicationRecord
+  enum :role, { gm: 0, player: 1, sub_keeper: 2 }, validate: true
+
+  belongs_to :play_session
+  belongs_to :person
+
+  validates :person_id, uniqueness: { scope: :play_session_id }
```
karia, line 7: [should-fix] **The same person twice in one submission 500s rather than showing a validation error.**

The uniqueness validator queries the database, so two *new* participations in a single nested submission both pass validation; the unique index then rejects the second INSERT. `ActiveRecord::RecordNotUnique` is not a `RecordNotFound`, so the new `rescue_from` in `ApplicationController` does not catch it.

Reproduced through the controller:

```
post manage_play_sessions_path, params: { play_session: { scenario_id: ..., participations_attributes: [
  { person_id: a.id, role: "gm" }, { person_id: a.id, role: "player" } ] } }
=> ActiveRecord::RecordNotUnique
```

This is an easy GM mistake — the person select in `_participation_row` lists everyone and nothing stops you picking the same name on two rows. Add a model-level check across the in-memory collection (a `validate` on `PlaySession` over `participations.reject(&:marked_for_destruction?)`, grouped by `person_id`) so the form comes back 422 with a message.

app/helpers/play_sessions_helper.rb
```diff
@@ -0,0 +1,16 @@
+module PlaySessionsHelper
+  def play_session_schedule(session)
+    return "日程未定" if session.played_on.blank?
+
+    date = l(session.played_on, format: :long)
+    session.started_at.present? ? "#{date} #{l(session.started_at, format: :short)}" : date
```
karia: [should-fix] **The schedule renders a bogus date: `2026年5月1日 2000-01-01 20:00`.**

`started_at` is a `time` column, so Rails hands back a Time on the dummy date 2000-01-01, and this repo's `ja.time.formats.short` is `"%Y-%m-%d %H:%M"` (config/locales/ja.yml:16, pre-existing). Verified:

```
helper.play_session_schedule(PlaySession.new(played_on: Date.new(2026,5,1), started_at: "20:00"))
=> "2026年5月1日 2000-01-01 20:00"
```

It shows up on the public list, the detail page, the scenario history and the manage list — every screen this PR adds. No spec asserts the rendered schedule text, which is why it slipped through: `spec/requests/play_sessions_spec.rb` only checks the scenario title, names and URLs.

Use a time-only format (`session.started_at.strftime("%H:%M")`, or add `ja.time.formats.hour_minute`) rather than the shared `:short`, and assert the string in a helper or request spec.

app/controllers/manage/play_sessions_controller.rb
```diff
@@ -0,0 +1,56 @@
+module Manage
+  class PlaySessionsController < BaseController
+    before_action :set_play_session, only: %i[edit update destroy]
+
+    def index
+      authorize PlaySession
```
karia: [should-fix] **This `authorize` grants everyone; the all-sessions listing is gated only by `require_editor`.**

`authorize PlaySession` calls `PlaySessionPolicy#index?`, which is `person.present?`. Verified: `PlaySessionPolicy.new(create(:person), PlaySession).index? => true`. So the one explicit authorization guarding the screen that deliberately lists *every* session regardless of visibility passes for any signed-in Person, and the real gate is the inherited `before_action :require_editor`.

That is a single point of failure for exactly the disclosure ADR-0001 warns about, and it reads as if a check is happening when none is. Prefer `authorize PlaySession, :create?` (or a dedicated `manage_index?`/`Manage::PlaySessionPolicy`) so the policy layer independently enforces editor-only, and add a request spec that a plain player gets 404 here — the current "answers 404 to a person with no role" spec would still pass with this line's predicate wide open.

app/controllers/manage/play_sessions_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+      authorize @play_session
+    end
+
+    def create
+      @play_session = authorize PlaySession.new(play_session_params)
+
+      if @play_session.save
+        redirect_to manage_play_sessions_path, notice: "セッションを登録しました"
+      else
+        @play_sessions = PlaySession.includes(:scenario).newest_first
```
karia: [should-fix] **N+1 on the create-failure re-render: `participations: :person` is dropped here but `index.html.erb` needs it.**

Line 8 preloads `participations: :person`; this one does not, yet `index.html.erb:15` calls `session.participations.map { |p| p.person.display_name }` for every row. Measured with 5 sessions × 2 participants:

```
GET  /manage/play_sessions        →  9 queries
POST /manage/play_sessions (422)  → 24 queries, 13 of them against "people"
```

Just reuse line 8's includes. Worth extracting to a private method so the two cannot drift again.

(Separately: `_participation_row` issues `Person.all` per rendered row, so an edit form with many participants re-queries the whole people table once per row. Hoisting it into a local in `_form` would fix that too.)

app/models/play_session.rb
```diff
@@ -0,0 +1,16 @@
+class PlaySession < ApplicationRecord
+  enum :status, { scheduled: 0, played: 1, cancelled: 2 }, validate: true
+
+  belongs_to :scenario
+  has_many :participations, -> { order(:position, :id) }, dependent: :destroy, inverse_of: :play_session
+  has_many :people, through: :participations
+
+  accepts_nested_attributes_for :participations, allow_destroy: true, reject_if: :all_blank
```
karia: [should-fix] **`reject_if: :all_blank` never fires for rows this form produces.**

`:all_blank` rejects a row only when *every* attribute except `_destroy` is blank. `_participation_row` renders the role select with no `include_blank`, so a freshly-added empty row always submits `role="gm"` and is never blank. Reproduced:

```
participations_attributes: [{ person_id: "", role: "gm", character_name: "", character_sheet_url: "", _destroy: "0" }]
→ participations.size = 1, valid? false, ["Participations personを入力してください"]
POST /manage/play_sessions → 422
```

So clicking 参加者を足す and then submitting without filling the row fails the whole save instead of being ignored, which is the case `reject_if` was presumably there for. Deleting the option entirely also leaves all 199 specs green, confirming nothing exercises it.

Either add `include_blank` to the role select, or replace with `reject_if: ->(attrs) { attrs["person_id"].blank? }`, and add a spec for the empty-row submission.

app/models/play_session.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  validates :recording_url, http_url: true
+
+  # 日付が無い回を末尾に固定する。データベース既定の NULL の並びに任せない。
+  scope :newest_first, -> { order(Arel.sql("played_on DESC NULLS LAST"), started_at: :desc, id: :desc) }
```
karia: [optional] **`started_at` sorts NULLS FIRST, the opposite of the intent stated one line above.**

Postgres defaults to NULLS FIRST for `DESC`, and only `played_on` was given an explicit `NULLS LAST`. So on a shared date a time-undecided session floats above the ones with times:

```
same day, started_at [20:00, nil, 13:00] → newest_first → [nil, 20:00, 13:00]
```

The comment says undated sessions are pinned to the end rather than left to the database default; the time component still is. `Arel.sql("started_at DESC NULLS LAST")` would make the two consistent. The ordering spec only varies `played_on`, so this is unverified either way today.

app/models/play_session.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def gm = participations.find { |p| p.gm? }&.person
```
karia: [optional] **`gm` is dead code.**

Nothing calls it — the views use `participation_role_label` and iterate participations directly. Rewriting it as `def gm = nil` leaves all 199 specs passing. Drop it, or use it in the list views where the GM is currently indistinguishable from the players in the joined name string.

app/policies/play_session_policy.rb
```diff
@@ -0,0 +1,31 @@
...
+          .where(person_id: visible_person_ids).arel.exists
+      )
+    end
+
+    private
+      def visible_person_ids
+        peers = GroupMembership.where(group_id: GroupMembership.where(person_id: person.id).select(:group_id))
+          .select(:person_id)
+
+        Person.where(id: peers).or(Person.where(id: person.id)).select(:id)
```
karia: [optional] **This reimplements `Person.sharing_a_group_with`, and the extra `Person` hop drags a `default_scope` ORDER BY into the subquery.**

`Person.sharing_a_group_with` exists with a comment saying "Phase 3 のセッション可視性がこの述語を使う" — that promise is now unkept, so the same peer-derivation lives in two places. Whichever you keep, the other should go.

The round trip through `Person` also has no effect other than pulling in `default_scope { order(:display_name) }`. Generated SQL:

```sql
person_id IN (SELECT people.id FROM people
  WHERE (people.id IN (SELECT group_memberships.person_id ...) OR people.id = 57)
  ORDER BY people.display_name ASC)
```

`peers` is already an id relation; `Participation.where(person_id: peers).or(Participation.where(person_id: person.id))` — or a `person_id IN (...) OR person_id = ?` inside the EXISTS — avoids the hop and the ordering. Behaviour is correct as written; this is purely tidying.

spec/requests/manage/play_sessions_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      expect(PlaySession.count).to eq(0)
+    end
+
+    it "lists every session regardless of who played, so a GM can maintain them" do
+      other = create(:play_session)
+      other.participations.create!(person: create(:person), role: :gm)
+
+      get manage_play_sessions_path
+
+      expect(response.body).to include(other.scenario.title)
```
karia: [should-fix] **This spec passes for the wrong reason and cannot detect the thing it names.**

`other.scenario.title` also appears in the `<option>` list of the scenario `collection_select` that `_form` renders at the top of `index.html.erb`, because that select is populated with `Scenario.order(:title)`. Measured on this exact fixture: the title occurs **twice** in the body, once inside an `<option>`.

So swapping line 8 of the controller to `policy_scope(PlaySession)...` — i.e. reintroducing the reader-facing scope on the maintenance screen, exactly the regression this spec exists to catch — leaves all 199 specs green. I verified that mutant survives.

Assert on something only the list can produce: the participant's `display_name` (the GM shares no group with them), or `edit_manage_play_session_path(other)`.

spec/requests/manage/play_sessions_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+
+    it "updates a session" do
+      session = create(:play_session, scenario:, status: :scheduled)
+
+      patch manage_play_session_path(session), params: { play_session: { status: "played" } }
+
+      expect(session.reload).to be_played
+    end
```
karia, line 162: [should-fix] **`destroy` and `edit` are entirely untested, and the write policies are only half-covered.**

Surviving mutants against the full 199-example suite:

| mutation | result |
|---|---|
| delete `authorize @play_session` from `destroy` | 199 pass |
| delete `authorize @play_session` from `edit` | 199 pass |
| `def destroy? = true` | 199 pass |
| `def create? = true` | 199 pass |
| `allow_destroy: true` → `false` | 199 pass |
| drop `-> { order(:position, :id) }` from `participations` | 199 pass |
| stop rendering `note` on the detail page | 199 pass |

(`update? = true` *is* killed, by the policy spec — `create?`/`destroy?` are simply not asserted there.)

The area is safe today because `require_editor` in `Manage::BaseController` catches all of it — I confirmed a plain player gets 404 on both `POST` and `DELETE`. But per ADR-0001 the policy is meant to be the enforcement point, and right now nothing holds it to that for two of the three write actions. Suggest: a `DELETE` request spec (asserting the row and its participations go), a non-editor `DELETE`/`POST` 404 spec, `create?`/`destroy?` added to the existing "lets GMs and admins write" example, a `_destroy` round-trip spec, and one assertion that `note` renders on the detail page.

karia (commented):
I tried hard to break `PlaySessionPolicy::Scope` and could not: I ran 26 mutants against the full suite, and every visibility mutation I could think of (anonymous sees all, admin bypass removed, self dropped from `visible_person_ids`, peers dropped, `show?` forced true, `index?` forced true, the controller or the history partial bypassing `policy_scope`) was killed by a spec. `note` does not reach an anonymous visitor, the sitemap only carries scenarios, and denied detail requests are a bodiless 404. The scope itself is sound.

What I did find is a cluster of problems in the manage area and the display layer, all reproduced by execution: two paths that 500 on ordinary GM mistakes, a schedule string that renders `2026年5月1日 2000-01-01 20:00`, an inert `reject_if`, an N+1 on the create-failure re-render, and one spec that passes for the wrong reason. Details inline.

app/views/manage/play_sessions/edit.html.erb
karia: Confirmed and fixed in 24f761f. Reproduced first: blanking the scenario select on update raised `ActionView::Template::Error` from `@play_session.scenario.title`.

The heading now reads `@play_session.scenario&.title || "セッション"`, and there is a spec that clears the scenario on update and expects 422 with the record unchanged.

app/models/participation.rb
karia: Confirmed and fixed in 24f761f. Reproduced `ActiveRecord::RecordNotUnique`.

Added `PlaySession#participants_are_distinct`, which checks the person ids of the in-memory rows (skipping any marked for destruction) before the DB constraint can fire. The constraint stays as the backstop.

Verified the spec kills the mutant: removing the validation drops the suite to 1 failure.

app/helpers/play_sessions_helper.rb
karia: Confirmed and fixed in 24f761f. Good catch — this was on every page the PR adds.

The cause is `l(..., format: :short)` resolving to this repo's `ja.time.formats.short`, which carries a date, applied to a `time` column whose date part is the 2000-01-01 placeholder. Now formats with `strftime("%H:%M")`.

Added a spec asserting the page contains `2026年5月1日 20:00` and does not contain `2000-01-01`. Restoring `l(:short)` fails the suite.

app/controllers/manage/play_sessions_controller.rb
karia: Fixed in 24f761f, and the same weakness existed in `Manage::ScenariosController` — `ScenarioPolicy#index?` is `true` for everyone, so that one was even wider.

Both policies now have `manage?`, defined as `editor?`, and the manage controllers authorize against it. `require_editor` stays as the outer gate; this is the second layer.

Added policy-level specs, because a request spec cannot see the difference while `require_editor` is in front. Weakening `manage?` to `person.present?` now fails the suite.

app/controllers/manage/play_sessions_controller.rb
karia: Fixed in 24f761f. Both the index and the create-failure re-render go through one `maintained_sessions` method carrying `includes(:scenario, participations: :person)`, and `_participation_row` now takes its list from a memoised `people_for_select` helper instead of `Person.all` per row.

Measured after the change: the create-failure re-render is 9 queries, down from the 24 you observed.

app/models/play_session.rb
karia: Confirmed and fixed in 24f761f. You are right that `:all_blank` could never fire — the role select had no blank option, so an empty row always carried `role="gm"`.

Two changes: the role select now has `include_blank`, and `reject_if` tests `person_id` directly rather than relying on the whole row being blank. A row is only meaningful once a person is chosen.

Added a spec submitting one filled row and one empty row and expecting a single participation.

app/models/play_session.rb
karia: Fixed in 24f761f. `started_at` now uses `NULLS LAST` too, matching the stated intent one line above.

Added a spec: two sessions on the same day, one with 20:00 and one with no time, must come back in that order. Reverting to `started_at: :desc` fails it.

app/models/play_session.rb
karia: Fixed in 24f761f — removed. It was left over from an earlier draft of the detail view.

app/policies/play_session_policy.rb
karia: Fixed in 24f761f, by deleting `Person.sharing_a_group_with` rather than calling it.

It had no other caller, its comment promised a Phase 3 use that never happened, and keeping both would leave two definitions of the same rule — which is exactly what ADR-0001 says to avoid for this particular rule. The scope now derives the peer ids from `GroupMembership` directly, with no `Person` hop, so the `default_scope` ordering no longer leaks into the subquery either.

Its spec went with it.

spec/requests/manage/play_sessions_spec.rb
karia: Confirmed and fixed in 24f761f. You are right that the scenario title also appears in the form's option list, so the assertion could not fail.

The spec now asserts the presence of `edit_manage_play_session_path(other)`, which only appears for a row that is actually listed. Re-scoping the manage index with `policy_scope` now fails it.

spec/requests/manage/play_sessions_spec.rb
karia: Fixed in 24f761f. Added specs for `edit`, `update` failure, `destroy`, `_destroy` on a participation, the duplicate-participant case, the empty-row case, and that the note reaches the edit form.

The manage spec file goes from 7 to 15 examples; the suite from 199 to 211.

One of your surviving mutants I chose to leave uncovered by a request spec: deleting `authorize` from `destroy`. `require_editor` makes it unobservable through HTTP. The policy-level spec added for `manage?` covers the equivalent ground at the layer where the difference is visible.

PR #17, opened 2026-08-10
Phase 4: profiles, aliases, favourites and the spoiler guard

karia:
## What

Implements [Phase 4](docs/plans/2026-08-09-phase4-profile-and-favorites.md), the last of the planned phases.

- Member list and profile pages, signed-in only
- A person edits their own profile; an admin can edit anyone's
- Aliases, with a note of where each name is used
- Favourites, toggled without a full page load
- The spoiler guard, filling the second empty partial Phase 1 left

## The spoiler guard

`ScenarioPolicy#show_preparation_note?` is the single point that decides, and the template omits the text server-side rather than hiding it. Specs assert the string is absent from the body for an anonymous visitor, for a member who has not pressed, and for a different member who has.

The press is recorded against the Person, so signing out and back in — or using another device — leaves it open. There is a spec for that.

## A privilege path I closed

`PersonPolicy#update?` now allows the person themselves, which is the point of the phase. That would have let a GM open `/manage/people/:id` for their *own* Person and set `group_ids`, and group membership decides who can read which sessions.

The manage screens now authorize against `manage?` (admin only) on every action, so the wider `update?` only applies to the profile screen, which does not accept `group_ids` at all.

## Verification

- [x] `bin/rspec` — 257 examples, 0 failures
- [x] `prek run --all-files`
- [x] A system spec walks sign-in, profile edit, spoiler reveal and favourite in one pass


karia (commented):
Third-party review. I ran the suite (257 green), mutation-tested the new specs, and probed the app with request specs. Good news first: the two claims hold. The spoiler guard is genuinely server-side (breaking `show_preparation_note?` to `person.present?` fails 4 examples, and swapping the partial to `reveal_preparation_note?` fails 3), and the `/manage/people` lockdown holds against a GM (index/edit/update/create/destroy all answer 404) with no route to group membership from the profile screen: `group_ids`, `roles` and `group_memberships_attributes` are all dropped, and a `person_aliases_attributes[][id]` pointing at someone else's alias raises RecordNotFound. What I did find: one member-triggerable 500 on a shared page via the new self-service icon upload, an N+1 on the member list, a cascade delete that quietly slipped in, and three mutations that survive the new specs.

app/controllers/people_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  private
+    def set_person
+      @person = policy_scope(Person).find(params[:id])
+    end
+
+    # グループ所属はここでは受け取らない。管理画面（管理者のみ）で扱う。
+    def person_params
+      params.expect(person: [ :display_name, :x_account, :icon,
```
karia, line 34: [blocking] **Any member can upload a non-image icon and 500 the member list for everyone.**

`:icon` is now writable by any signed-in member (before this PR only admins could attach one via `/manage/people`), and there is no content-type or size validation. I attached a `text/plain` file through `PATCH /people/:id` and it saved fine (302), after which both shared pages blow up for every viewer:

```
upload status=302 attached=true type=text/plain
member list RAISED ActionView::Template::Error: Can't transform blob with ID=20 and content_type=text/plain
profile     RAISED ActionView::Template::Error: Can't transform blob with ID=20 and content_type=text/plain
```

`accept: "image/*"` on the file field is client-side only. `icon.attached?` is true but the blob is not variable, so `icon.variant(:thumb)` raises `ActiveStorage::InvariableError` in `app/views/people/index.html.erb:12` and `app/views/people/show.html.erb:7` — one bad upload takes down `/people` for the whole group, and only a console fixes it.

Suggest validating the attachment on `Person` (content type in an allowlist plus a byte-size cap) and additionally guarding the views with `member.icon.variable?` so an already-stored bad blob degrades to the empty circle instead of a 500. Worth a request spec that uploads a `.txt` and expects a validation error rather than a redirect.

app/controllers/people_controller.rb
```diff
@@ -0,0 +1,37 @@
+class PeopleController < ApplicationController
+  before_action :set_person, only: %i[show edit update]
+
+  def index
+    authorize Person
+    @people = policy_scope(Person).includes(:person_aliases, :groups)
```
karia: [should-fix] **N+1 on the member list: two queries per member for the icon.**

`includes(:person_aliases, :groups)` covers the aliases and groups but not the attachment. Measured with an `sql.active_record` subscriber:

```
4 members  => 13 queries
9 members  => 23 queries   # +2 per member: active_storage_attachments, then active_storage_blobs
```

`.with_attached_icon` folds both into two constant queries.

app/models/scenario.rb
```diff
@@ -17,6 +17,9 @@ class Scenario < ApplicationRecord
   has_many :game_systems, through: :scenario_game_systems
   has_many :scenario_authors, dependent: :destroy
   has_many :authors, through: :scenario_authors
+  has_many :favorites, dependent: :destroy
+  has_many :spoiler_reveals, dependent: :destroy
+  has_many :play_sessions, dependent: :destroy
```
karia: [should-fix] **`dependent: :destroy` on `play_sessions` turns a blocked delete into a silent cascade, and it isn't part of this phase.**

`favorites` and `spoiler_reveals` need the cascade (new FKs, otherwise a scenario can never be deleted). `play_sessions` does not — that FK already existed on main, where deleting a scenario with sessions was a hard stop:

```
# with this line removed
ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: update or delete on table "scenarios"
  violates foreign key constraint "fk_rails_c93681f206" on table "play_sessions"

# with this line
destroy status=302 sessions=0 participations=0
```

`ScenarioPolicy#destroy? = editor?`, so a GM clicking 削除 on `/manage/scenarios` now silently wipes every play session and participation for that scenario — including sessions that GM cannot even see under `PlaySessionPolicy::Scope`. No spec covers it. Either drop the line (keep the FK violation) or add `dependent: :restrict_with_error` and surface a message, but please not a bare cascade smuggled into the profile phase.

spec/requests/favorites_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  it "is closed to an account that is not linked to a person" do
+    sign_in_as create(:user, person: nil)
+
+    post scenario_favorite_path(scenario)
+
+    expect(response).to have_http_status(:not_found)
+    expect(Favorite.count).to eq(0)
+  end
+
+  it "adds and removes a favourite for the signed-in member" do
```
karia, line 22: [should-fix] **Nothing pins the favourite to `current_person` — this mutation survives the whole suite.**

I replaced the controller line with:

```ruby
Person.find(params[:person_id] || current_person.id).favorites.find_or_create_by!(scenario: @scenario)
```

217 examples, 0 failures. The current code is correct, but the property "one member cannot create a favourite for another" is unguarded, so a future refactor that reads a person from params slips through. Add an example that posts `person_id: other.id` and asserts `other.favorites` stays empty while `person.favorites` gains the row.

spec/requests/people_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+      person.person_aliases.create!(name: "べつの名前", context: "とあるサーバ")
+      sign_in_as other
+
+      get person_path(person)
+
+      expect(response.body).to include("べつの名前", "とあるサーバ")
+    end
+  end
+
+  describe "editing" do
```
karia, line 41: [should-fix] **`GET /people/:id/edit` has no authorization coverage — this mutation survives too.**

Every example here exercises `PATCH`. Changing `PeopleController#edit` to `authorize @person, :show?` (i.e. any signed-in member can load anyone's edit form) leaves 217 examples, 0 failures. `verify_authorized` is still satisfied, so nothing catches the wrong policy method.

Add `get edit_person_path(person)` as `other` expecting 404, and as the owner and an admin expecting 200.

spec/system/profile_and_spoiler_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    OmniAuth.config.test_mode = true
+    OmniAuth.config.mock_auth[:google_oauth2] = OmniAuth::AuthHash.new(
+      provider: "google_oauth2", uid: user.google_uid, info: { email: user.email }
+    )
+    visit root_path
+    click_button "Google でログイン"
+
+    visit edit_person_path(person)
+    fill_in "person[x_account]", with: "karia"
+    fill_in "person[person_aliases_attributes][0][name]", with: "べつの名前" rescue nil
```
karia: [should-fix] **`rescue nil` means the alias half of this flow is never exercised.**

The person is created with no aliases, so `form.fields_for :person_aliases` renders no persisted row and the only alias inputs on the page are the `NEW_RECORD` template ones. `fill_in "person[person_aliases_attributes][0][name]"` therefore always raises `Capybara::ElementNotFound`, the `rescue nil` swallows it, and no assertion ever checks the alias was saved — so the one end-to-end test of nested aliases tests nothing. A `rescue nil` in a spec also hides every future breakage at that line.

Under `rack_test` the 別名を足す button is inert (Stimulus doesn't run), so either seed `person.person_aliases.create!` first and edit the rendered row, or drop the line and cover the add/remove path in the request spec — where it would also be worth asserting the real form shape `person_aliases_attributes: { "0" => {...} }`, since the request spec only ever sends the array shape while the form sends a hash of indexes.

app/controllers/favorites_controller.rb
```diff
@@ -0,0 +1,27 @@
+class FavoritesController < ApplicationController
+  before_action :set_scenario
+
+  def create
+    authorize @scenario, :favourite?
+    current_person.favorites.find_or_create_by!(scenario: @scenario)
```
karia: [optional] **`find_or_create_by!` under a double-submit answers 422 or 500 rather than being idempotent.**

A plain double press is fine (I confirmed the second POST is a no-op, 200, count stays 1). But the check-then-insert has two racy outcomes: if the second request's uniqueness validation sees the first row it raises `RecordInvalid` → 422; if both validations run before either insert, the unique index raises `RecordNotUnique`, which `ApplicationController` does not rescue → 500. I forced the first case:

```
race favourite status=422
```

`create_or_find_by` plus a `rescue ActiveRecord::RecordNotUnique` (or just rescuing both and re-rendering the button) makes the endpoint honestly idempotent. Same shape in `SpoilerRevealsController#create`.

app/models/person.rb
```diff
@@ -17,8 +21,14 @@ class Person < ApplicationRecord
     define_method(:"#{role}?") { person_roles.any? { |r| r.name == role.to_s } }
   end
 
+  accepts_nested_attributes_for :person_aliases, allow_destroy: true, reject_if: ->(attrs) { attrs["name"].blank? }
```
karia: [optional] **`reject_if` also silences edits to existing aliases, so clearing a name looks like a successful save.**

The lambda runs for persisted rows too, not just new ones. Blanking an existing alias's name in the form reports success and changes nothing:

```
blank rename status=302 name=keep
```

The user's intent ("remove this name") is only honoured through the 削除 checkbox. Scoping the guard to new records — `->(attrs) { attrs["id"].blank? && attrs["name"].blank? }` — lets `PersonAlias`'s presence validation surface the error instead of dropping the change.

app/controllers/people_controller.rb
karia: Confirmed and fixed in 7bb7da0. Reproduced exactly as described: a `text/plain` upload succeeded, and the next `GET /people` raised from `icon.variant(:thumb)`. One member could take down the shared list for everyone.

Added `active_storage_validations` and validated both attachments — icon `[:png, :jpeg, :gif, :webp]` under 5 MB, jacket the same types under 10 MB. Chose the gem over a hand-rolled check because attachment validation has more edge cases than it looks (declared vs sniffed content type).

Spec asserts the upload is rejected with 422, nothing is attached, and `GET /people` still answers 200 afterwards. Removing the validation fails 3 examples.

app/controllers/people_controller.rb
karia: Fixed in 7bb7da0 with `.with_attached_icon`. Measured after: 19 queries for 10 members, and it no longer grows by 2 per member.

app/models/scenario.rb
karia: Confirmed and fixed in 7bb7da0. This was a regression I introduced in this PR — I added the association while wiring up favourites and reveals, without thinking about what `dependent: :destroy` means for it.

Changed to `dependent: :restrict_with_error`, and the controller now reports the error instead of calling `destroy!`. A GM deleting a scenario with sessions gets a message rather than silently destroying play records they may not even be able to see.

Two specs: deletion refused while sessions exist, deletion allowed when none do. Restoring `:destroy` fails one.

spec/requests/favorites_spec.rb
karia: Fixed in 7bb7da0. Added a spec that signs in as one member, posts with `person_id` set to another member, and asserts the favourite lands on the signed-in member and not the victim.

Your mutant now fails.

spec/requests/people_spec.rb
karia: Fixed in 7bb7da0. Added two specs for `GET /people/:id/edit`: 404 for someone else, 200 for the person themselves. Authorizing `edit` with `:show?` now fails.

spec/system/profile_and_spoiler_spec.rb
karia: Fixed in 7bb7da0. You are right that `rescue nil` meant the alias flow was never exercised — the `rescue` was a shortcut I should not have left in.

The spec now creates a persisted alias first, asserts the field appears with its current value, renames it, and asserts the new name shows on the profile while the old one is gone.

The array-vs-hash shape difference you noted is real. The nested attributes accept both, and the system spec now drives the actual form, so the hash shape is covered there while the request specs keep using arrays for brevity.

app/controllers/favorites_controller.rb
karia: Fixed in 7bb7da0. `create!` wrapped so that both `RecordInvalid` and `RecordNotUnique` fall through to rendering the button in its pressed state. Pressing twice is not an error — the outcome the member asked for is already true.

app/models/person.rb
karia: Fixed in 7bb7da0. `reject_if` now only skips rows with no `id`, so blanking an existing alias reaches the validation and comes back as 422 instead of a silent no-op.

PR #21, opened 2026-08-10
Make the player role implicit for every member

### LINKED ISSUE #19: ユーザーは全員プレイヤー権限に所属する
ユーザー管理画面で権限「プレイヤー」のチェックボックスがオンオフ可能になっているが、全員がプレイヤーなので以下の通り変更する

1. チェックボックスはオンで固定し管理者含め変更を不可とする
2. 作成されたユーザーは既存含め全員「プレイヤー」権限を持っているものとする
3. 閲覧できる画面は、現在の「権限なし」同等とする（ログインユーザーであれば見られるセッション画面等は引き続きみられる／非ログインユーザーがセッション画面を見られないのは現状維持）

karia:
Closes #19

## Approach

The role is now implicit rather than stored. A row that every person carries holds no information, and it can drift — a Person created outside the form would silently lack it. `Person#player?` returns true unconditionally, `player` is gone from `PersonRole::ROLES`, and a migration deletes the rows that exist.

Submitting `roles: ["player"]` is ignored rather than rejected, so an old form or a stale tab cannot produce a 422.

## Against the three requirements

1. The checkbox renders ticked and `disabled`, with a line saying it cannot be removed. There is no form field behind it, so nothing can post it.
2. Everyone is a player, existing rows included — the migration drops the stored rows and the predicate answers for all.
3. Visibility is unchanged. Verified by request: anonymous and unlinked accounts get 404 on `/people` and `/play_sessions`; a member with no stored roles gets 200 on both and 404 on `/manage/*`. Specs assert the player role grants nothing an unroled person lacks.

## Verification

- [x] `bin/rspec` — 279 examples, 0 failures
- [x] `prek run --all-files`
- [x] The member list shows プレイヤー for everyone

## Note

`down` cannot restore who held a player row, since everyone did. It re-inserts one for every person, which is the state this issue describes.

### DIFF
--- a/app/helpers/people_helper.rb
+++ b/app/helpers/people_helper.rb
@@ -1,0 +1,6 @@
+module PeopleHelper
+  # プレイヤーは保存しないが、全員が持つものとして必ず表示する。
+  def person_role_labels(person)
+    [ t("people.roles.player") ] + person.roles.map { |role| t("people.roles.#{role}") }
+  end
+end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -15,6 +15,7 @@ class Person < ApplicationRecord
   validates :display_name, presence: true
   validates :icon, content_type: [ :png, :jpeg, :gif, :webp ], size: { less_than: 5.megabytes }
   validate :keeps_at_least_one_admin
+  validate :roles_are_known
 
   default_scope { order(:display_name) }
 
@@ -29,7 +30,8 @@ class Person < ApplicationRecord
   accepts_nested_attributes_for :person_aliases, allow_destroy: true,
     reject_if: ->(attrs) { attrs["id"].blank? && attrs["name"].blank? }
 
-  def roles = person_roles.map(&:name)
+  # 未知の値が入った行は名前を引けない。表示側でロケール全体を出さないよう落とす。
+  def roles = person_roles.map(&:name).compact
 
   def revealed?(scenario) = spoiler_reveals.exists?(scenario_id: scenario.id)
 
@@ -38,14 +40,22 @@ class Person < ApplicationRecord
   def self.admins = joins(:person_roles).where(person_roles: { name: PersonRole.names[:admin] })
 
   # has_many への代入は永続レコードだと即座に DB へ反映される。検証で見るため代入前の状態を残す。
-  # フォームが player を送ってきても落とす。保存しない権限なので無視してよい。
+  # 古いフォームが送ってくる player だけを黙って落とす。それ以外の未知の値は検証エラーにする。
+  # 不正な行をそのまま代入すると RecordNotSaved で 500 になるため、ここで取り分ける。
   def roles=(names)
     @roles_before_assignment = roles if persisted? && !defined?(@roles_before_assignment)
-    assignable = Array(names).compact_blank.uniq & PersonRole::ROLES.keys.map(&:to_s)
-    self.person_roles = assignable.map { |name| PersonRole.new(name:) }
+    submitted = Array(names).compact_blank.uniq - %w[player]
+    @unknown_roles = submitted - PersonRole::ROLES.keys.map(&:to_s)
+    self.person_roles = (submitted - @unknown_roles).map { |name| PersonRole.new(name:) }
   end
 
   private
+    def roles_are_known
+      return if @unknown_roles.blank?
+
+      errors.add(:base, "知らない権限です: #{@unknown_roles.join(", ")}")
+    end
+
     # 全員が管理者でなくなると、rake タスク以外に復旧手段が無くなる。
     def keeps_at_least_one_admin
       return unless defined?(@roles_before_assignment)
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -22,6 +22,9 @@ class Person < ApplicationRecord
     define_method(:"#{role}?") { person_roles.any? { |r| r.name == role.to_s } }
   end
 
+  # Person であることがそのままプレイヤーであることを表す。付け外しはできない。
+  def player? = true
+
   # 既存行の名前を空にしたときは無視せず検証に落とす。新規の空行だけ捨てる。
   accepts_nested_attributes_for :person_aliases, allow_destroy: true,
     reject_if: ->(attrs) { attrs["id"].blank? && attrs["name"].blank? }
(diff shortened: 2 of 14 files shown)


spec/requests/manage/player_role_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    expect(response).to redirect_to(manage_people_path)
+    expect(person.reload.roles).to eq([ "gm" ])
+  end
+
+  it "lists プレイヤー for everyone on the member index" do
+    person
+
+    get manage_people_path
+
+    expect(response.body).to include("プレイヤー")
```
karia: [should-fix] **This assertion is vacuous — it passes even when the helper stops emitting the player label.**

I mutated `person_role_labels` to `person.roles.map { ... }` (dropping the prepended player label) and the whole suite stayed green at 279/0. The reason is that `manage/people/index.html.erb` renders `_form` at the top, and that partial already contains `プレイヤー` twice (the disabled checkbox label and the new explanatory paragraph), so `response.body` matches regardless of what the list renders.

`person_role_labels` is the only new production code with no effective coverage. Either assert against the list row specifically (e.g. match `権限: プレイヤー` for the person, which the form does not produce) or add `spec/helpers/people_helper_spec.rb` asserting `person_role_labels(create(:person))` == `["プレイヤー"]` — `spec/helpers/` already exists for `scenarios_helper`.

app/models/person.rb
```diff
@@ -35,9 +38,11 @@ def favourite?(scenario) = favorites.exists?(scenario_id: scenario.id)
   def self.admins = joins(:person_roles).where(person_roles: { name: PersonRole.names[:admin] })
 
   # has_many への代入は永続レコードだと即座に DB へ反映される。検証で見るため代入前の状態を残す。
+  # フォームが player を送ってきても落とす。保存しない権限なので無視してよい。
   def roles=(names)
     @roles_before_assignment = roles if persisted? && !defined?(@roles_before_assignment)
-    self.person_roles = Array(names).compact_blank.uniq.map { |name| PersonRole.new(name:) }
+    assignable = Array(names).compact_blank.uniq & PersonRole::ROLES.keys.map(&:to_s)
```
karia: [optional] **The intersection swallows every unknown role, not just `player`, turning a validation failure into a silent no-op.**

On `main`, an unknown name built a `PersonRole` that failed the enum validation and autosave rejected the whole update. Now:

```
p = Person.create!(display_name: "x", roles: %w[gm])
p.update(roles: %w[owner])  # => true, errors: []
p.reload.roles              # => []
```

Same for a case typo (`%w[GM]` yields `[]`). The admin-lockout guard still fires, so there is no escalation path, but a future rename of a `ROLES` key that misses the form would silently wipe roles instead of erroring. Dropping only the value this PR is about keeps that loud:

```ruby
assignable = Array(names).compact_blank.uniq - %w[player]
```

app/helpers/people_helper.rb
```diff
@@ -0,0 +1,6 @@
+module PeopleHelper
+  # プレイヤーは保存しないが、全員が持つものとして必ず表示する。
+  def person_role_labels(person)
+    [ t("people.roles.player") ] + person.roles.map { |role| t("people.roles.#{role}") }
```
karia, line 4: [optional] **A role row whose stored integer is not in `ROLES` makes `roles` contain `nil`, and `t("people.roles.")` then renders the entire translation hash into the page.**

The migration's `down` recreates exactly that state — I ran it on a dev DB:

```
db:migrate:down  =>  person_roles: {0=>1, 2=>2}
Person.all.map(&:roles)  =>  [["admin", nil], [nil]]
person_role_labels(person)  =>  ["プレイヤー", {admin: "管理者", gm: "GM", player: "プレイヤー"}]
```

So a `db:rollback` that is not paired with a code revert leaves the member list printing `権限: プレイヤー、{admin: "管理者", gm: "GM", player: "プレイヤー"}`. Not a crash and not reachable after a normal deploy, but `person.roles.compact` here costs nothing and keeps a stray row from leaking the locale file into the page. (The `down` itself is fine otherwise: it round-trips, and the unique index is not at risk because `up` deletes every player row first.)

karia (commented):
Third-party review. The change looks sound: nothing on `main` ever branched on `player?` (`git grep player? origin/main -- app` is empty), so requirement 3 holds trivially — the visible surface is unchanged for everyone. I verified the migration end to end against a dev DB (`db:migrate:down` then `:up` round-trips; `2` was indeed the `player` enum value and no other code used it), confirmed the enum validation genuinely blocks a stored player row (`person_roles.create!(name: "player")` and `name: 2` both raise `RecordInvalid`), and that no `person[...]` field is posted for the disabled checkbox (it renders as `name="person_roles_player_display"`, outside `params.expect(person: ...)`). I also mutation-tested the new specs: flipping `player?` to `false` kills 6 examples, restoring `player: 2` to `ROLES` kills 3, and dropping the filter in `roles=` kills 2 — all three mutants die. Three non-blocking notes below, one of which is a mutant that survives.

spec/requests/manage/player_role_spec.rb
karia: Confirmed and fixed. You are right that the assertion could not fail — the index renders `_form`, which carries the word twice.

The spec now asserts the row format itself, `権限: プレイヤー`, and a second one asserts `権限: プレイヤー、GM` for a person who also holds a stored role. Stripping the label from `person_role_labels` now fails 2 examples.

app/models/person.rb
karia: Fixed, though not with `- %w[player]` alone. That form raises rather than validating.

Assigning an invalid record through `self.person_roles =` on a persisted row saves immediately and raises `ActiveRecord::RecordNotSaved`, so `p.update(roles: %w[owner])` would have been a 500 rather than a rejection. Reproduced before changing it.

`roles=` now drops `player` silently, sets aside anything else it does not recognise, and a `roles_are_known` validation turns that into an error. An unknown name comes back as a failed update with a message, and a stale form carrying `player` still succeeds.

Specs cover both. Reverting to the silent `&` filter fails 1 example.

app/helpers/people_helper.rb
karia: Fixed at the source rather than in the helper: `Person#roles` now compacts, so every reader is covered rather than just this one.

Added a spec that inserts a `person_roles` row with an unknown integer via raw SQL and asserts `roles` returns only the known one. Removing the `compact` fails it.

PR #23, opened 2026-08-10
Hide the recommendation and add a table view to the scenario list

### LINKED ISSUE #18: オススメ度をフロントから非表示にする
「オススメ度」は、シナリオの編集画面以外では非表示にする
ソート順序を管理するためのパラメーターとして内部では持ち続けるが、今後立て付けを変更する

### LINKED ISSUE #20: シナリオ一覧を表形式とジャケット優先表示で切り替え可能にする
シナリオ一覧が現在ジャケットを強調した表示デザイン（ジャケット優先表示とよぶ）になっていますが、従来スプレッドシートで一覧性を重視して見ていたので変えたい

1. 表形式での一覧表示を実装し、ジャケット優先表示を切り替えられるようにする
2. トップページでは、表形式をデフォルトに変更する
3. 表形式では、「シナリオ名」「作者」「システム」「人数」「目安時間」を表示する。クリックするとシナリオ詳細に飛べる。また「入手先」も表示し、こちらはリンク先は実際の販売サイトURLとする。いずれも、DBで空の項目は表示上空欄としてよい。

karia:
Closes #18
Closes #20

Both issues touch the same screen, so they are here together as two commits.

## #18 — the recommendation leaves the front

Gone from the list, the jacket view and the scenario page; still on the edit form. It now drives the default ordering instead: best first, unrated last, ties by title. The site description mentioned it too, and no longer does.

## #20 — table by default, jackets on request

`?view=table` (the default) and `?view=gallery`, chosen because it keeps the URL shareable, behaves the same signed out, and leaves caching alone — the alternative of remembering the choice in a cookie makes one URL render differently per visitor. An unknown value falls back to the table.

The table carries シナリオ名 / 作者 / システム / 人数 / 目安時間 / 入手先. The title links to the scenario; the purchase label links to the shop. Empty columns render empty rather than 未設定, which the list previously showed — the table reads better without a column of placeholder text.

## Verification

- [x] `bin/rspec` — 293 examples, 0 failures
- [x] `prek run --all-files`
- [x] Rendered and inspected: a fully populated row comes out as `["見本シナリオ", "見本作者", "CoC 7版", "3人", "6時間〜8時間", "BOOTH"]`, an empty scenario as blanks, and no `★` anywhere

### DIFF
--- a/app/helpers/scenarios_helper.rb
+++ b/app/helpers/scenarios_helper.rb
@@ -5,24 +5,27 @@ module ScenariosHelper
     name == current ? "font-bold text-ink no-underline" : "text-seal"
   end
 
-  # 一覧では、値が無い欄は空欄のままにする（issue #20）。
-  def blank_to_empty(label) = label == UNSET ? "" : label
 
-  def player_count_label(scenario)
+  # 値が無ければ nil。一覧は空欄に、詳細は「未設定」に落とす。
+  def player_count_value(scenario)
     numeric = bounded_label(scenario.player_count_min, scenario.player_count_max) { |n| "#{n}人" }
 
-    [ numeric, scenario.player_count_note ].compact_blank.join(" ").presence || UNSET
+    [ numeric, scenario.player_count_note ].compact_blank.join(" ").presence
   end
 
-  def duration_label(scenario)
+  def player_count_label(scenario) = player_count_value(scenario) || UNSET
+
+  def duration_value(scenario)
     min = scenario.duration_min_minutes
     max = scenario.duration_max_minutes
-    return UNSET if min.blank? && max.blank?
+    return nil if min.blank? && max.blank?
 
     # 2 時間未満は分のままのほうが元データに近く、範囲でも単位が揃う。
     unit = [ min, max ].compact.max < 120 ? :minutes : :hours
     bounded_label(min, max) { |n| humanized_minutes(n, unit:) }
   end
+
+  def duration_label(scenario) = duration_value(scenario) || UNSET
 
   def character_sheet_deadline_label(scenario)
     return scenario.character_sheet_deadline_note if scenario.character_sheet_deadline_note.present?
--- a/app/helpers/scenarios_helper.rb
+++ b/app/helpers/scenarios_helper.rb
@@ -1,6 +1,13 @@
 module ScenariosHelper
   UNSET = "未設定".freeze
 
+  def list_view_class(name, current)
+    name == current ? "font-bold text-ink no-underline" : "text-seal"
+  end
+
+  # 一覧では、値が無い欄は空欄のままにする（issue #20）。
+  def blank_to_empty(label) = label == UNSET ? "" : label
+
   def player_count_label(scenario)
     numeric = bounded_label(scenario.player_count_min, scenario.player_count_max) { |n| "#{n}人" }
 
@@ -15,13 +22,6 @@ module ScenariosHelper
     # 2 時間未満は分のままのほうが元データに近く、範囲でも単位が揃う。
     unit = [ min, max ].compact.max < 120 ? :minutes : :hours
     bounded_label(min, max) { |n| humanized_minutes(n, unit:) }
-  end
-
-  def recommendation_label(scenario)
-    return "回したことない" unless scenario.gm_experienced
-    return "未評価" if scenario.recommendation.blank?
-
-    "★" * scenario.recommendation
   end
 
   def character_sheet_deadline_label(scenario)
--- a/app/views/scenarios/_gallery.html.erb
+++ b/app/views/scenarios/_gallery.html.erb
@@ -1,0 +1,33 @@
+<ul class="mt-8 grid gap-px border border-rule bg-rule sm:grid-cols-2 lg:grid-cols-3">
+  <% scenarios.each do |scenario| %>
+    <li class="bg-surface">
+      <%= link_to scenario_path(scenario), class: "group flex h-full flex-col gap-3 p-4 no-underline text-ink" do %>
+        <div class="aspect-3/4 overflow-hidden bg-paper">
+          <% if scenario.jacket.attached? %>
+            <%= image_tag scenario.jacket.variant(:thumb), alt: "", loading: "lazy",
+                  class: "h-full w-full object-cover" %>
+          <% else %>
+            <span class="flex h-full items-center justify-center font-display text-sm text-muted">
+              画像なし
+            </span>
+          <% end %>
+        </div>
+
+        <h2 class="font-display text-lg leading-snug group-hover:text-seal">
+          <%= scenario.title %>
+        </h2>
+
+        <p class="text-xs text-muted">
+          <%= scenario.game_systems.map(&:name).join("／") %>
+        </p>
+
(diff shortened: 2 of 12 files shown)


spec/requests/scenario_list_views_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+    it "still orders the list, best first" do
+      create(:scenario, title: "低評価", recommendation: 1)
+      create(:scenario, title: "未評価", recommendation: nil)
+
+      get root_path
+
+      body = response.body
+      expect(body.index("見本シナリオ")).to be < body.index("低評価")
```
karia, line 106: [should-fix] **This ordering assertion passes even with the ordering removed.**

I replaced the scope body with `all` (`scope :recommended_first, -> { all }`) and ran `rspec spec/requests/scenario_list_views_spec.rb spec/system/browsing_scenarios_spec.rb`: 14 examples, 0 failures. The three rows are created in exactly the order the test expects, and for a table this small Postgres hands them back in insertion order, so the assertion never exercises `recommendation DESC NULLS LAST`.

Creating them worst-first fixes it, e.g. `未評価` (nil), then `低評価` (1), and let `見本シナリオ` (5) — created by the outer `let!` — sort up from last place. Worth adding a tie case too: two scenarios with the same recommendation and titles in reverse alphabetical creation order, so the `:title` secondary key is covered as well. (I confirmed both orderings do work in a probe spec; it is only the test that is blind.)

karia (commented):
Reviewed against #18 and #20 and ran the suite locally. The implementation itself holds up: `render "scenarios/#{@view}"` is safely gated by the `VIEWS` allowlist (array, hash, empty and `../` traversal values all fall back to the table with a 200), the recommendation is gone from the table, gallery, scenario page, JSON-LD, meta description and sitemap, and the ordering behaves correctly at runtime (NULLS LAST plus the title tiebreak). Full suite, RuboCop, erb_lint and Brakeman are green. My findings are mostly about the tests: two of the key guarantees survive being deliberately broken.

spec/requests/scenario_list_views_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+      expect(response.body).to include("<table")
+    end
+  end
+
+  describe "the recommendation" do
+    it "is absent from the list" do
+      get root_path
+
+      expect(response.body).not_to include("★")
```
karia: [should-fix] **These assertions guard the star rendering, not the value, so a leak of the raw recommendation survives.**

I added `data-recommendation="<%= scenario.recommendation %>"` plus `<%= scenario.recommendation %>` to the 入手先 cell in `_table.html.erb` and the whole suite stayed green (14 examples, 0 failures). `★` and `おすすめ度` only exist because the old helper produced them; any future reintroduction as a number, a data attribute, a Stimulus value or a sort-key class goes unnoticed — which is precisely the regression #18 asks to prevent.

Adding `expect(response.body).not_to match(/recommendation/i)` alongside these catches the attribute/class/JSON shapes. Applying it to the gallery and scenario-page examples below covers the rest.

app/helpers/scenarios_helper.rb
```diff
@@ -1,6 +1,13 @@
 module ScenariosHelper
   UNSET = "未設定".freeze
 
+  def list_view_class(name, current)
+    name == current ? "font-bold text-ink no-underline" : "text-seal"
+  end
+
+  # 一覧では、値が無い欄は空欄のままにする（issue #20）。
+  def blank_to_empty(label) = label == UNSET ? "" : label
```
karia: [optional] **`blank_to_empty` parses a formatted label back into a blank, and can swallow a real value.**

A scenario with no `player_count_min`/`max` whose `player_count_note` is literally `未設定` renders an empty 人数 cell — I verified this in a probe request spec. It is also inconsistent: with `player_count_min: 3` and the same note, the cell renders `3人 未設定`, so the same note is dropped in one case and kept in the other.

The cause is that the helper compares against `UNSET`, which is a presentation string rather than a signal of absence. Having `player_count_label`/`duration_label` return `nil` when there is nothing to say, and letting the caller supply `UNSET` (show page) or `""` (list), removes the round-trip. Contrived input, hence optional, but the coupling is the kind that quietly breaks when a label is reworded.

app/views/scenarios/index.html.erb
```diff
@@ -1,45 +1,18 @@
...
-            <%= image_tag scenario.jacket.variant(:thumb), alt: "", loading: "lazy",
-                  class: "h-full w-full object-cover" %>
-          <% else %>
-            <span class="flex h-full items-center justify-center font-display text-sm text-muted">
-              画像なし
-            </span>
-          <% end %>
-        </div>
+  <nav class="flex items-center gap-3 text-xs" aria-label="表示の切り替え">
+    <%= link_to "表形式", root_path(view: "table"), class: list_view_class("table", @view) %>
```
karia: [optional] **The active view is conveyed only by `font-bold`.**

`list_view_class` gives the current link boldness and no underline, but the markup carries nothing programmatic — rendered output is `<a class="font-bold text-ink no-underline" href="/?view=table">表形式</a>`. Screen reader users, and anyone whose stylesheet does not load, cannot tell which view is active, and both entries are still self-links.

`aria-current="page"` on the active link (or rendering it as a `<span>` instead of a link) resolves it.

app/views/scenarios/index.html.erb
```diff
@@ -1,45 +1,18 @@
...
-          <dt class="text-muted">評価</dt>
-          <dd class="<%= "text-seal" if scenario.recommendation.present? %>">
-            <%= recommendation_label(scenario) %>
-          </dd>
-        </dl>
-      <% end %>
-    </li>
-  <% end %>
-</ul>
+<%= render "scenarios/#{@view}", scenarios: @scenarios %>
```
karia: [optional] **The empty state now shows a headers-only table above the "no scenarios" message.**

With zero scenarios the partial still renders the bordered container, `<thead>` with all six column headings and an empty `<tbody>`, and the `まだシナリオが登録されていません。` panel appears underneath it (verified by rendering `root_path` with an empty table). Reads as a broken table rather than an empty catalog.

Moving the render into the `else` of the emptiness check, i.e. `<% if @scenarios.empty? %> … <% else %><%= render … %><% end %>`, avoids it.

app/views/scenarios/_gallery.html.erb
```diff
@@ -0,0 +1,33 @@
...
+            </span>
+          <% end %>
+        </div>
+
+        <h2 class="font-display text-lg leading-snug group-hover:text-seal">
+          <%= scenario.title %>
+        </h2>
+
+        <p class="text-xs text-muted">
+          <%= scenario.game_systems.map(&:name).join("／") %>
```
karia: [optional] **The gallery lost its `システム未設定` fallback in the move to a partial.**

`main` had `join("／").presence || "システム未設定"` here (`app/views/scenarios/index.html.erb:26` on `main`); the extracted partial drops the fallback, so a scenario with no game system now renders an empty `<p>` — confirmed by requesting `?view=gallery` for such a scenario. #20's "空欄としてよい" clause is about the new table's columns, so this looks like collateral rather than an intended change. No test covers either behaviour, so restore the fallback or drop it deliberately.

PR #26, opened 2026-08-10
Document setup, contributing and the migration procedure

### LINKED ISSUE #25: ドキュメント整備
以下のドキュメントを整備する

README.mdに初期セットアップ、コントリビュート方法を明示
CLAUDE.mdにアーキテクチャや実装方針、テスト実行方法、DBマイグレーション方法などを記載

いずれも簡潔に記載するように注意。ソースを読めばわかることは記載しない。howを記載するように心掛ける。

karia:
Closes #25

## README

Aimed at someone who has just cloned the repo: what to install, how to get it running, and what has to pass before a PR. Everything in it was run in a clean worktree while writing it.

The one thing worth arguing over is the note about authorization — a contributor who does not know that the public and private areas share a process will not think to write the spec for it.

## CLAUDE.md

Kept as the working document rather than a second README, so it links back for setup and covers what is easy to get wrong instead:

- The migration procedure, including that production migrations run as a separate Job **before** the image is swapped, and why the reverse order breaks
- What to do when the local database misbehaves, and how to spot leftover rows in the test database — this cost me a confusing failure earlier today
- Where each layer lives, so a change lands in the right place

## Also fixed: `bin/dev`

Documenting it forced me to run it, and it died instantly. `bin/rails tailwindcss:watch` exits straight away without a TTY, and foreman then tears down Puma with it. `tailwindcss:watch[always]` fixes it. Writing a setup guide around a broken command was not an option.

## Verification

- [x] Every command in both documents executed: `mise install`, `bin/setup --skip-server`, `bin/dev`, `bin/rspec`, single-file `bin/rspec`, `prek run --all-files`, `db:seed`, the migration commands, the reset recipe, the leftover-rows check
- [x] `bin/rspec` — 294 examples, 0 failures


CLAUDE.md
```diff
@@ -1,44 +1,72 @@
...
+```bash
+bin/rails generate migration AddFooToBar foo:string
+bin/rails db:migrate          # 開発 DB に当て、db/schema.rb が更新される
+bin/rails db:test:prepare     # テスト DB に schema.rb を流し直す
+```
+
+`db/schema.rb` は必ず commit する。テストは migration ではなく schema.rb から DB を作る。
+
+**本番のマイグレーションはアプリの起動から切り離してある。** コンテナの entrypoint では走らない。
+`yuno04-k3s` の `apps/trpg-catalog/migrate-job.yaml` を Job として流し、そのあとに Deployment のイメージを差し替える。
```
karia: [should-fix] **The migration procedure omits the two steps that make it actually work.**

I checked this against `yuno04-k3s/apps/trpg-catalog/`. Following the text as written would migrate nothing:

1. `migrate-job.yaml` pins its own image digest (currently `sha-a55be156...@sha256:47015...`), independently of `deployment.yaml`. You must bump the Job's image to the new build *first*; otherwise you run the old code's migrations and then swap in a new image against an unmigrated schema — exactly the failure this section warns about.
2. The Job uses `generateName`, so it has to be run with `kubectl create -f`, not `kubectl apply -f`. The infra repo's own README calls this out explicitly.

Also worth being precise: the Job runs `./bin/rails db:prepare`, not `db:migrate`, which matters because the Solid Cache/Queue/Cable databases are created separately by CloudNativePG.

Suggested: "`yuno04-k3s` の `apps/trpg-catalog/migrate-job.yaml` の image を新しいビルドに差し替えてから `kubectl create -f` で流し（`generateName` なので `apply` では流せない）、そのあとで Deployment のイメージを差し替える。"

Minor aside: naming a path inside the private infra repo sits slightly against this file's own rule on line 60. The repo name was already here on `main`, so this is not new exposure — just flagging the tension.

README.md
```diff
@@ -5,9 +5,61 @@ TRPG のシナリオとセッションのカタログサイト。
...
+## 変更を出す
 
-- [ADR](docs/adr/) — アプリケーションの設計判断
-- [実装計画](docs/plans/) — フェーズごとの作業手順
+```bash
+bin/rspec                 # テスト
+prek run --all-files      # RuboCop / erb_lint / Brakeman / gitleaks
+```
+
+この2つが通ることが、PR を出す前提。CI でも同じものが走る。
```
karia: [should-fix] **"CI でも同じものが走る" is not true, and the command that *is* equivalent goes undocumented.**

`bin/rspec` + `prek run --all-files` is a strict subset of `.github/workflows/ci.yml`, which also runs `bin/bundler-audit`, `bin/importmap audit`, and a full production Docker image build. A contributor can have both documented commands green and still get a red CI.

The repo already ships the right answer: `bin/ci` (driven by `config/ci.rb`) runs setup, RuboCop, erb_lint, bundler-audit, importmap audit, Brakeman and RSpec — the CI matrix minus the image build. Neither document mentions it. I ran it here: passes in 27s.

Suggested: document `bin/ci` as the pre-PR command, and soften the claim to something like "CI はこれに加えて本番イメージのビルドを確かめる".

README.md
```diff
@@ -5,9 +5,61 @@ TRPG のシナリオとセッションのカタログサイト。
...
+docker compose up -d      # PostgreSQL と MinIO
+bin/setup                 # 依存の取得、DB 作成、開発サーバの起動まで
+```
+
+`bin/setup` はサーバまで起動する。DB の用意だけしたいときは `bin/setup --skip-server`。
+
+`.env` は編集しなくても動く。既定値が `compose.yaml` に合わせてある。
+Google 認証を手元で試すときだけ `GOOGLE_CLIENT_ID` と `GOOGLE_CLIENT_SECRET` が要る。
+
+シナリオが 0 件だと画面が寂しいので、見本を入れておくとよい。
```
karia: [should-fix] **On a fresh clone the database is not empty at this point — `bin/setup` has already seeded it.**

`bin/setup` runs `bin/rails db:prepare`, and `ActiveRecord::Tasks::DatabaseTasks#prepare_all` calls `load_seed` whenever it had to initialize a database (activerecord-8.1.3.1, `database_tasks.rb:174-206`). So the first-time reader who just followed the block above already has the 2 example scenarios in, and the premise "シナリオが 0 件だと画面が寂しいので" doesn't describe their state.

Where `bin/rails db:seed` genuinely earns its place is re-running after a schema reset, or loading real data via `SCENARIOS_SEED_FILE`. Suggest reframing along those lines, or noting that `bin/setup` already did this on a fresh database.

CLAUDE.md
```diff
@@ -1,44 +1,72 @@
...
+
+## 手元の DB が壊れたら
+
+```bash
+docker compose down -v && docker compose up -d
+bin/rails db:prepare && bin/rails db:test:prepare
+```
+
+テストが理由なく落ちるときは、テスト DB に残留データが無いか疑う。
+`RAILS_ENV=test bin/rails runner 'puts Person.count'` が 0 でなければ、過去の実行が残している。
```
karia: [should-fix] **This points the reader at a cause that this suite structurally cannot have.**

`spec/rails_helper.rb:17` sets `config.use_transactional_fixtures = true`, so every example rolls back — including the two system specs, since Rails shares the connection with the browser thread. I ran the full suite and then the exact command in this doc: `Person=0 Scenario=0 PlaySession=0`. Leftover data is not a plausible first suspect here, so "テストが理由なく落ちるときはまずこれを疑え" will cost someone a detour.

The genuinely useful failure mode in this repo is a stale test schema after pulling a migration — which `maintain_test_schema!` mostly handles, but not when `db/schema.rb` and the test database drift. Suggest replacing these two lines with that, or cutting them: the "手元の DB が壊れたら" recipe above already covers the recovery.

CLAUDE.md
```diff
@@ -1,44 +1,72 @@
...
 
-```
-mise install
-cp .env.example .env
-docker compose up -d
-bin/rails db:prepare
-```
+## 構成
+
+Rails 8 の SSR モノリス。設計判断は [ADR-0001](docs/adr/0001-application-architecture.md)。
```
karia: [optional] **Duplicated verbatim with README.md:53-54 — the issue asks to cut exactly this.**

Both files now open their architecture section with "Rails 8 の SSR モノリス" plus a link to ADR-0001. Line 3 of this file already delegates to the README, so this line pays nothing.

Cut it and let the table stand on its own — the table is the part of this section that isn't available anywhere else, and it's the strongest thing in the file: the four-row layer breakdown plus the `User`/`Person` distinction on lines 16-17 is real orientation that reading the source would take a while to reconstruct.

README.md
```diff
@@ -5,9 +5,61 @@ TRPG のシナリオとセッションのカタログサイト。
...
+必要なのは [mise](https://mise.jdx.dev/) と Docker の2つ。Ruby と Node は mise が入れる。
+
+```bash
+mise install
+cp .env.example .env
+docker compose up -d      # PostgreSQL と MinIO
+bin/setup                 # 依存の取得、DB 作成、開発サーバの起動まで
+```
+
+`bin/setup` はサーバまで起動する。DB の用意だけしたいときは `bin/setup --skip-server`。
```
karia: [optional] **First sentence restates the inline comment three lines up.**

Line 18 already says `bin/setup  # 依存の取得、DB 作成、開発サーバの起動まで`. "`bin/setup` はサーバまで起動する。" adds nothing on top of it.

Keep only the half that carries new information: "DB の用意だけしたいときは `bin/setup --skip-server`。" (verified working.)

karia (commented):
Third-party review. I followed both documents end to end on a clean checkout: `bin/setup --skip-server`, `bin/rails db:seed` (twice, confirmed idempotent), `bin/rails db:test:prepare`, `bin/rspec` (294 examples, 0 failures), `bin/rspec spec/requests/scenarios_spec.rb:42`, `prek run --all-files` (all 8 hooks pass), and the `RAILS_ENV=test` runner one-liner. Every command works as written, and I cross-checked the stated conventions against `app/policies/`, `ApplicationController`, `Manage::BaseController` and `lib/tasks/admin.rake` — the 404-not-403 rule, `pundit_user = current_person`, `Person#player? = true` and `ScenarioPolicy#show_preparation_note?` all match the code. The `Procfile.dev` fix is correct and I confirmed it empirically: without a TTY, `bin/rails tailwindcss:watch` exits with status 0 in ~1s (which takes foreman down with it), while `tailwindcss:watch[always]` was still running at a 25s timeout; `always` only appends the `-w always` flag in tailwindcss-rails 4.6.0, so an interactive terminal is unaffected and CI never touches this path (`db:test:prepare` is enhanced with `tailwindcss:build`, not `watch`). Six findings below — two factual gaps in the newcomer path, one incomplete production procedure, one piece of misleading debugging advice, and two cuts the issue explicitly asks for. No credentials or non-public identifiers introduced.

CLAUDE.md
karia: Confirmed against the infra repo and fixed. All three points were true, and the omissions were the ones that actually bite.

The section is now a three-step list: bump the digest on **both** the Job and the Deployment, `kubectl create -f` the Job (it uses `generateName`, so `apply` cannot create it), confirm it finished, then `apply` the Deployment. It also says the Job runs `db:prepare`.

The independent digest is the part I would have got wrong myself in a hurry, which is exactly why it belongs here.

README.md
karia: Fixed. The claim was false: `ci.yml` also runs `bin/bundler-audit`, `bin/importmap audit` and a production image build.

The README now points at `bin/ci`, which I had not noticed the repo already ships. Ran it: 27s, all green. `bin/rspec` and `prek run --all-files` are still mentioned as the quick loop while writing code.

README.md
karia: Fixed. You are right — `db:prepare` calls `load_seed` when it creates the database, so a fresh clone already has the two example scenarios. The text described a state the reader is never in.

Now says the seed comes in with `bin/setup`, and `bin/rails db:seed` is only for putting it back.

CLAUDE.md
karia: Fixed. Reworded so leftover data is what you check *after* ruling out the ordinary causes, not first, and it says outright that each example rolls back.

Keeping a mention rather than cutting it: I hit this exact case earlier today and spent a while on it, because a stale row made a legitimate failure look like a passing suite. Rare, but expensive when it happens.

CLAUDE.md
karia: Fixed — the heading and the two duplicated lines are gone. The section is now just the table, under 「どこに何があるか」.

README.md
karia: Fixed, along with the sentence above it. The setup block now reads as one paragraph rather than restating the comments in the snippet.

While re-running the commands for these fixes I also noticed `db/seeds.rb` printing `already initialized constant SEED_ATTRIBUTES` whenever it was loaded twice, which the seeds spec does. Changed to a method; the warning is gone.

PR #74, opened 2026-08-11
Add alias management for systems, authors, and members

### LINKED ISSUE #46: 別名機能の実装
以下の要素に対し、別名機能を実装する。

* システム
* 作者
* メンバー

### 別名機能の詳細

以下の様に定義する

1. 上記各要素には、名前を1個以上複数設定できる
2. それぞれの名前について「表示」「非表示」の属性が選択できる
4. 名前のうちどれか一つを選択して「表示名」に設定しなければならない。サイト上の一覧や詳細など、他の画面からの引用では「表示名」が利用される。「表示名」に設定した名前は、強制で「表示」扱いとなる（「非表示」設定の名前は「表示名」に設定できない）。

### 権限

* システム・作者はGM権限ユーザーおよび管理者が設定可
* メンバーは自分自身および管理者が設定可

karia:
## What

Add visible and hidden aliases for game systems, authors, and members, with promotion of a visible alias to the display name.

## Why

Support the naming rules described in #46 while keeping one required display name for references across the site.

## Verification

- [x] `bin/rspec`
- [x] `prek run --all-files`
- [x] `bin/ci`

## Related

Closes #46

### DIFF
--- a/app/javascript/controllers/nested_form_controller.js
+++ b/app/javascript/controllers/nested_form_controller.js
@@ -2,11 +2,31 @@ import { Controller } from "@hotwired/stimulus"
 
 // 入手先と配信リンクの行を増やす。増やした行は保存時にまとめて送られる。
 export default class extends Controller {
-  static targets = ["template", "anchor"]
+  static targets = ["template", "anchor", "aliasName", "aliasSelect"]
   static values = { token: { type: String, default: "NEW_RECORD" } }
+
+  connect() {
+    this.syncAliasOptions()
+  }
 
   add() {
     const html = this.templateTarget.innerHTML.replaceAll(this.tokenValue, new Date().getTime())
     this.anchorTarget.insertAdjacentHTML("beforebegin", html)
+    this.syncAliasOptions()
+  }
+
+  syncAliasOptions() {
+    if (!this.hasAliasSelectTarget) return
+
+    const selected = this.aliasSelectTarget.value
+    const options = [new Option(this.aliasSelectTarget.dataset.baseLabel, "")]
+
+    this.aliasNameTargets.forEach((input) => {
+      const name = input.value.trim()
+      if (name !== "") options.push(new Option(name, input.dataset.selectionKey))
+    })
+
+    this.aliasSelectTarget.replaceChildren(...options)
+    if (options.some((option) => option.value === selected)) this.aliasSelectTarget.value = selected
   }
 }
--- a/app/javascript/controllers/nested_form_controller.js
+++ b/app/javascript/controllers/nested_form_controller.js
@@ -2,7 +2,7 @@ import { Controller } from "@hotwired/stimulus"
 
 // 入手先と配信リンクの行を増やす。増やした行は保存時にまとめて送られる。
 export default class extends Controller {
-  static targets = ["template", "anchor", "aliasName", "aliasSelect"]
+  static targets = ["template", "anchor", "aliasRow", "aliasSelect"]
   static values = { token: { type: String, default: "NEW_RECORD" } }
 
   connect() {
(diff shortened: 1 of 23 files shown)


app/models/concerns/has_aliases.rb
```diff
@@ -0,0 +1,39 @@
+module HasAliases
+  extend ActiveSupport::Concern
+
+  included do
+    attr_accessor :display_alias_id
+
+    accepts_nested_attributes_for :aliases, allow_destroy: true,
+      reject_if: ->(attrs) { attrs["id"].blank? && attrs["name"].blank? }
+
+    before_validation :promote_selected_alias
```
karia, line 10: The display-name swap runs in `before_validation` and is not idempotent. If another validation fails, the rendered values are already swapped, and resubmitting can swap them back and save the unintended name. Please make retries preserve the same meaning and cover validation failure followed by resubmission.

app/views/shared/_alias_row.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="grid gap-2 sm:grid-cols-[auto_1fr_auto_auto]">
+  <% if row.object.persisted? && row.object.visible? %>
+    <label class="inline-flex items-center gap-1 text-xs text-muted">
+      <%= form.radio_button :display_alias_id, row.object.id %> 表示名
+    </label>
+  <% else %>
+    <span class="self-center text-xs text-muted">別名</span>
+  <% end %>
+  <%= row.text_field :name, placeholder: "別名", class: "border border-rule bg-paper px-3 py-2 text-sm" %>
```
karia: The previous member alias form exposed `PersonAlias#context`, but the shared row no longer renders it. This regresses creating and editing the Discord/server context. Please retain that field for person aliases and add coverage.

app/views/shared/_alias_row.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="grid gap-2 sm:grid-cols-[auto_1fr_auto_auto]">
+  <% if row.object.persisted? && row.object.visible? %>
```
karia: Only persisted, already-visible aliases get a display-name radio. A newly added alias or one made visible in the same edit cannot be selected until a second save. Please support selecting it in one submission and cover the flow.

app/models/concerns/has_aliases.rb
karia: Fixed in 0717ae9. Failed validations now restore the pre-promotion names before rendering, so retrying keeps the selected promotion direction. Added shared model coverage for restoration after an unrelated validation failure.

app/views/shared/_alias_row.html.erb
karia: Fixed in 0717ae9. The shared alias row renders the context field for PersonAlias records, and the request spec now covers updating it.

app/views/shared/_alias_row.html.erb
karia: Fixed in 0717ae9. Alias rows now submit a stable selection key, including dynamically added rows, so a new visible alias can become the display name in the same update. Added request coverage.

karia:
Deployed the locally built PR image to the k3s verification environment.

- Migration `AddAliases` completed; schema version is `20260812000000`
- Web and worker are both `1/1 Ready` on the PR image
- External `/up` returned HTTP 200 five consecutive times
- No rollout-time 5xx or error logs

Infrastructure tracking: karia/yuno04-k3s#63

app/javascript/controllers/nested_form_controller.js
```diff
@@ -2,11 +2,31 @@ import { Controller } from "@hotwired/stimulus"
...
+  }
+
+  syncAliasOptions() {
+    if (!this.hasAliasSelectTarget) return
+
+    const selected = this.aliasSelectTarget.value
+    const options = [new Option(this.aliasSelectTarget.dataset.baseLabel, "")]
+
+    this.aliasNameTargets.forEach((input) => {
+      const name = input.value.trim()
```
karia: The display-name select is built from alias names only, so an alias made hidden or marked for deletion remains selectable and is then rejected with 422. Please build options row-by-row from name, visible, and destroy state, and reset an invalid current selection.

karia: Fixed in b0c7cd8. The controller now reads each alias row as a unit and only offers nonblank aliases that are visible and not marked for deletion. Input and change events both refresh the select, and an invalid current selection falls back to the existing display name.

karia:
Updated the k3s verification deployment with the reviewed form-layout changes (`b0c7cd8`). Web and worker are both `1/1 Ready`; external `/up` returned HTTP 200 five consecutive times; no rollout-time 5xx or error logs were found.

PR #102, opened 2026-08-15
Fix the flaky sign-in race in the profile system spec

karia:
## What

Wait for the sign-in redirect to render before the next `visit` in `spec/system/profile_and_spoiler_spec.rb`.

## Why

This spec has been failing intermittently on `main` (`Unable to find field "person[x_account]"`). The failure screenshot from CI shows the browser sitting on the post-login root page, flash and all, at the moment the field was looked for.

Neither `click_button` nor `visit` waits for a navigation to finish, so the `visit` raced the OAuth round-trip and the late-arriving login redirect replaced the edit page. `editor_navigation_spec.rb` does the same sign-in but follows it with `click_button "メニュー"` — a retrying finder for an element that only renders when signed in — so it synchronises by accident and never flaked.

Only reproducible under the Chrome driver; locally `CHROME_BINARY` is unset, so system specs run on `:rack_test`, which is synchronous.

## Verification

- [x] `bin/rspec`
- [x] `prek run --all-files`

Reproduced against real Chrome (Chrome for Testing 152) by running the spec in a loop: **1 failure in 40 runs before, 0 in 150 runs after**.

### DIFF
--- a/spec/system/profile_and_spoiler_spec.rb
+++ b/spec/system/profile_and_spoiler_spec.rb
@@ -39,6 +39,8 @@ RSpec.describe "A member's own pages" do
 
     visit scenario_path(scenario)
     click_button "お気に入りに入れる"
+    # ボタンの差し替えを待たずに visit すると、飛びかけた POST ごと捨てられる。
+    expect(page).to have_button("お気に入りを外す")
 
     visit person_path(person)
     expect(page).to have_content("見本シナリオ")
--- a/spec/system/profile_and_spoiler_spec.rb
+++ b/spec/system/profile_and_spoiler_spec.rb
@@ -13,6 +13,9 @@ RSpec.describe "A member's own pages" do
     )
     visit root_path
     click_button "Googleでログイン"
+    # click_button も visit も遷移の完了を待たないため、ここで着地を待たないと
+    # 後から届いたログインのリダイレクトが次の visit を上書きする。
+    expect(page).to have_content("ログインしました")
 
     visit edit_person_path(person)
     fill_in "person[x_account]", with: "karia"


spec/system/profile_and_spoiler_spec.rb
```diff
@@ -13,6 +13,9 @@
     )
     visit root_path
     click_button "Googleでログイン"
+    # click_button も visit も遷移の完了を待たないため、ここで着地を待たないと
+    # 後から届いたログインのリダイレクトが次の visit を上書きする。
+    expect(page).to have_content("ログインしました")
```
karia, line 18: (anchored here because GitHub only accepts comments inside the diff — this is about **line 41-43** of the same file.)

The same unwaited-navigation race is still present further down:

```ruby
click_button "お気に入りに入れる"

visit person_path(person)
expect(page).to have_content("見本シナリオ")
```

That `button_to` submits through Turbo (`FavoritesController#create` answers with `turbo_stream.replace`), so `click_button` returns as soon as the click is dispatched, without waiting for the POST to finish. Under the Chrome driver the immediately following `visit` can abort the in-flight fetch before the server creates the favourite; the person page then renders no favourites and the spec fails with `expected to find text "見本シナリオ"` — the same class of flake this PR fixes one screen earlier.

Worth adding the same kind of landing check before the `visit`, e.g. `expect(page).to have_button("お気に入りを外す")`, which only appears once the turbo_stream replacement has been applied.

karia, line 18: Minor, but this now puts the whole sign-in round trip inside Capybara's wait window, and `Capybara.default_max_wait_time` is never raised anywhere (`spec/support/capybara.rb` only registers the driver), so it stays at the 2 s default.

On a cold CI container the first real request through Chrome pays for eager-load/asset work, and the OmniAuth POST → callback → redirect → render chain can plausibly exceed 2 s. In that case the flake does not disappear, it just moves here and reports as `expected to find text "ログインしました"` on the root page. Setting `Capybara.default_max_wait_time = 5` for the Chrome driver would make the new synchronisation point as robust as intended.

karia: 指摘のとおり、同じ機序のraceが残っていました。`FavoritesController#create` が `turbo_stream.replace` でボタンを差し替える作りであることを確認したうえで、`visit` の前に `expect(page).to have_button("お気に入りを外す")` を入れました。

karia: 今回は見送ります。OAuthの往復はOmniAuthのtest modeでプロセス内完結するためネットワークを介さず、2秒を超える要素は想定しにくいこと、また既存のアサーションもすべて同じ2秒に依存しており、今回追加した待ち合わせだけが特別に不利になるわけではないことが理由です。CIで実際にタイムアウト側に振れた観測が出た時点で、根拠を持って引き上げる方が適切と考えます。

PR #103, opened 2026-08-15
Confirm operations that strip your own roles

### LINKED ISSUE #98: 管理者としてログイン中に、自分自身の権限が降格するような操作で警告を表示する
管理者ユーザーが自分自身の権限が降格する操作ができてしまい、実行した瞬間に管理者向けページが表示できなくなる

このような操作パターンを一通り洗い出し、試行しようとした場合は警告を表示する

## 具体例

discordで管理者ユーザーとしてログインしているときに、管理者ユーザーのdiscord紐付けを解除すると、管理者権限が失われ404 Not Foundになる

* 現在：操作できる
* 変更後：操作時に管理者権限が失われる旨の警告を表示し、本当に操作継続するのかを確認する

discordで管理者ユーザーとしてログインしているときに、管理者ユーザーのgoogle紐付けを解除すると、引き続き操作ができる

* 現在：操作できる
* 変更後：ログインユーザーの管理者権限は失われず操作継続できるので、これは警告を出さなくてよい（現状維持）

karia:
## What

Ask for confirmation before an operation strips the signed-in person of a role they hold.

Three paths are covered, each only when it hits the account you are actually signed in with:

- `Manage::Users#update` — relinking or unlinking your own account
- `People#update` — unchecking `admin` or `gm` on your own person
- `People#destroy` — deleting your own person

Unlinking another provider of the same person, or touching someone else, goes through unchanged.

## Why

Closes #98. Unlinking your own Discord account while signed in through Discord dropped admin on the spot and the next page was a 404, with no way back except `bin/rails admin:grant`.

## Verification

- [x] `bin/rspec`
- [x] `prek run --all-files`

## Related

`People#destroy` skips `keeps_at_least_one_admin`, so the last admin can still be deleted. Left alone here.


app/views/people/edit.html.erb
```diff
@@ -4,6 +4,10 @@
 
 <%= form_with model: @person, class: "mt-6 space-y-6" do |form| %>
   <%= accessible_error_summary(@person) %>
+  <% if @lost_roles.present? %>
+    <%= render "shared/self_demotion_warning", roles: @lost_roles %>
+    <%= hidden_field_tag :confirm_self_demotion, true %>
```
karia: `confirm_self_demotion` is embedded in the form and stays there for every subsequent submit from that page, so the acknowledgement is not scoped to the change that was warned about.

Concrete scenario (verified against this branch): an `admin`+`gm` person unchecks only GM and submits. The warning lists GM alone (`管理者権限を失い` is absent). On the re-rendered form they change their mind and also uncheck 管理者, then submit. Because the hidden field is still present, `roles_lost_by` short-circuits and the request goes straight through — the person ends up with `roles == []`, having never been warned about losing admin, which is the lockout this PR exists to prevent.

`app/views/manage/users/edit.html.erb:10` has the same shape: after one warning the user can pick a different 紐づけ先 and it is applied without re-checking.

Binding the token to the change it acknowledged (e.g. carry the roles that were warned about, and re-warn when the new submission loses something not on that list) would close this.

app/controllers/people_controller.rb
```diff
@@ -36,6 +39,10 @@ def create
 
   def destroy
     authorize @person, :destroy?
+
+    @lost_roles = @person == current_person ? roles_lost_by([]) : []
```
karia, line 43: The confirmation added here turns the one unguarded lockout path into a one-click affordance, and the copy it shows is wrong for exactly that case.

Verified on this branch:

```ruby
admin = create(:person, roles: %w[admin])   # the only admin
sign_in_as admin
delete person_path(admin)                    # 422, warning says 「別の管理者に権限を付け直してもらう」
delete person_path(admin), params: { confirm_self_demotion: "1" }
# => 302, Person.admins.count == 0
```

`destroy` never runs `keeps_at_least_one_admin` (that validation is only reachable via `save`), so pressing 了解して続行 leaves the site with zero admins and no recovery except `bin/rails admin:grant`. Meanwhile `people.self_demotion.recovery` promises another admin can put the role back, which is false in the only scenario where it matters.

By contrast the `update` path is safe — the last admin gets the same warning but the confirmed submit is rejected 422 by the validation. Suggest guarding `destroy` with the same "is anyone else an admin" check (refuse rather than confirm), or at minimum swapping the recovery text when `Person.admins.where.not(id: @person.id)` is empty.

app/controllers/people_controller.rb
```diff
@@ -17,6 +17,9 @@ def edit
   def update
     authorize @person, :update?
 
+    @lost_roles = roles_lost_by_role_change
```
karia, line 20: The confirmation gate returns before `@person.update`, so a submission that is *both* a demotion and invalid never gets validated. The 422 page shows only the demotion warning and an empty `accessible_error_summary`.

Verified: `patch person_path(admin), params: { person: { display_name: "", roles: ["gm", ""] } }` returns 422 with no error text in the body. The user sees a warning about losing admin, clicks 了解して続行, and only then learns 表示名 is blank — and because the failed `update` rolls the whole thing back, the role change is discarded too, so both edits have to be redone.

Running validations first (`@person.valid?` on the assigned copy, or checking the demotion only once the record is otherwise savable) would let both problems surface in one round trip.

app/views/shared/_self_demotion_warning.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="border border-seal bg-surface p-4 text-sm text-seal" role="alert" tabindex="-1" autofocus>
+  <h2 class="font-bold"><%= t("people.self_demotion.title") %></h2>
```
karia, line 2: Two a11y problems with this block, given the repo's WCAG 2.2 AA / NN-g rule.

1. Heading order. On `people/show.html.erb` the warning is rendered at the top of `<article>`, above the page's `<h1>` (line 27 of that file). The rendered document therefore starts `h2` → `h1`, which fails the "headings in a meaningful sequence" expectation and makes the warning look like a subsection of nothing to a screen-reader user navigating by heading. Either drop this to a non-heading element with `aria-labelledby`, or move the block below the `<h1>` on the show page.

2. `role="alert"` plus `autofocus` double-announces. The alert is present at page load, not inserted dynamically, so the live region fires while focus is simultaneously moved into it — AT reads the contents twice. Pick one: keep `tabindex="-1"` + `autofocus` for focus management and drop `role="alert"`, or keep the live region and let focus stay put.

app/views/people/edit.html.erb
karia: Fixed in fd93b45. The token now carries the roles the warning actually named (`confirm_self_demotion="admin"`), and `roles_lost_by` subtracts only those, so re-widening the form after the warning triggers a fresh one. Pinned by "warns again when the form drops a role the warning did not cover" and "carries the acknowledged roles on the re-rendered form".

app/controllers/people_controller.rb
karia: Half taken. The copy was mine and it was false for a last admin, so `recovery` is now "権限を戻すには、別の管理者に付け直してもらうか、サーバで管理者を設定し直す必要があります。" — true whether or not another admin remains, and no branching.

Leaving the `keeps_at_least_one_admin` gap on `destroy` itself, though. It predates this PR, `destroy` has no UI entry point at all (`app/views/manage/people/` is dead — no `Manage::PeopleController`, no route), so reaching the confirm page takes a hand-made DELETE. Adding a `before_destroy` guard also means handling `RecordNotDestroyed` in the controller, which is a wider change than this issue asked for. Still noted under Related.

app/controllers/people_controller.rb
karia: Not changing this one. The outcome is right in both orders — the update is rejected and no role is dropped — so the cost is one extra round trip in the narrow case of an invalid field *and* a self-demotion in the same submit.

Validating before the gate would mean assigning the attributes first, and `roles` / `manual_group_ids` write to the DB on assignment. `update` gets away with it because `ActiveRecord::Persistence#update` wraps assign+save in a transaction that rolls back; the gate runs outside one, which is exactly why `render_role_change_warning` excludes those two. Buying the round trip back would mean reproducing that transaction around a path whose whole point is not to write yet.

app/views/shared/_self_demotion_warning.html.erb
karia: Heading order: fixed in fd93b45. On `people/show` the warning now renders after the header block that holds the `<h1>`, so the document reads h1 → h2. It is still the last thing on the page above the fold and carries the confirm button, so nothing is buried.

Double announcement: keeping `role="alert"` + `tabindex="-1"` + `autofocus` as is. Neither mechanism is reliable alone here — the response is a full render, and SRs vary on whether an alert present at load time is announced, while `autofocus` depends on Turbo re-applying it after a body swap. `accessible_error_summary` already pays the same cost for the same reason (it sets `role="alert"` and then calls `.focus()` from `error_summary_controller`), so this matches the pattern already in the codebase rather than introducing a second, weaker one.

PR #104, opened 2026-08-15
Implement deletion for members, accounts, sessions and scenarios

### LINKED ISSUE #99: 削除がない箇所に削除を実装する
削除機能がない要素に対して削除を実装する
少なくとも以下に削除ボタンがないので実装するが、他の要素に対しても横並びで確認し実装する

* メンバー
* アカウント
* セッション
* シナリオ

また、各詳細画面に「このXXを削除」ボタンを実装する

## 権限

* 各要素を削除できるのは、管理者メンバーのみ
* 管理者メンバーであっても以下を削除することはできない
  * 管理者メンバーそのもの（自分自身）
  * 管理者メンバーに紐付くアカウント

karia:
## What

Add the missing delete paths for members, accounts, sessions and scenarios, and put a delete button on every detail and list screen (scenarios, sessions, members, accounts, groups, game systems, authors).

Guard rails:

- An admin cannot delete themselves, nor the account bound to themselves.
- A member who has taken part in a session cannot be deleted; the reason is shown instead of a 500.
- Deletion permissions are otherwise unchanged: members and accounts stay admin-only, scenarios and sessions stay editor-level.

Also drops three `manage/` views that no route reaches and that pointed at non-existent `manage_person_path` / `manage_play_session_path`.

## Why

Closes #99. The `destroy` actions already existed for most resources but had no UI, and accounts had no delete path at all.

## Verification

- [x] `bin/rspec`
- [x] `prek run --all-files`


app/views/scenarios/_gallery.html.erb
```diff
@@ -32,7 +32,14 @@
           <dd><%= duration_value(scenario) %></dd>
         </dl>
       <% end %>
-      <% if policy(scenario).update? %><p class="px-4 pb-4 text-xs"><%= link_to "編集", edit_scenario_path(scenario), data: { turbo_frame: "_top" }, class: "text-seal" %></p><% end %>
+      <% if policy(scenario).update? || policy(scenario).destroy? %>
+        <p class="flex flex-wrap items-center gap-3 px-4 pb-4 text-xs">
```
karia: `button_to` renders a `<form>`, and `<form>` is one of the elements that implicitly closes an open `<p>` in the HTML5 parsing algorithm. So this markup does not render as written — the browser produces:

```html
<p class="flex flex-wrap items-center gap-3 px-4 pb-4 text-xs"><a>編集</a></p>
<form class="inline" ...><button>削除</button></form>
<p></p>
```

(verified with Nokogiri::HTML5, the spec-conformant parser)

Concretely: for a GM on the gallery view, the 削除 button escapes the paragraph, so it loses `px-4 pb-4` (flush against the card edge, no bottom padding), it no longer sits on the same flex row as 編集, and an empty `<p></p>` is appended inside the `<li>`. Before this PR the element only held a link, so the `<p>` was valid.

Swapping the `<p>` for a `<div>` fixes it.

```suggestion
        <div class="flex flex-wrap items-center gap-3 px-4 pb-4 text-xs">
```
(and the matching closing tag below)

app/views/people/index.html.erb
```diff
@@ -24,7 +24,14 @@
           <% end %>
         </span>
       <% end %>
-      <% if policy(member).update? %><p class="px-4 pb-4 text-xs"><%= link_to "編集", edit_person_path(member), class: "text-seal" %></p><% end %>
+      <% if policy(member).update? || policy(member).destroy? %>
+        <p class="flex flex-wrap items-center gap-3 px-4 pb-4 text-xs">
```
karia: Same problem as in `scenarios/_gallery.html.erb`: `button_to` emits a `<form>`, which implicitly closes an open `<p>` per the HTML5 parsing algorithm. An admin viewing the member list gets the 削除 button hoisted out of this paragraph — it loses `px-4 pb-4`, drops off the flex row shared with 編集, and leaves a stray empty `<p></p>` in the `<li>`.

```suggestion
        <div class="flex flex-wrap items-center gap-3 px-4 pb-4 text-xs">
```
(and change the matching `</p>` to `</div>`)

app/views/play_sessions/index.html.erb
```diff
@@ -15,7 +15,15 @@
           <%= session.participations.map { |p| "#{participation_role_label(p)}: #{p.person.display_name}" }.join("、") %>
         </span>
       <% end %>
-      <% if policy(session).update? %><p class="px-4 pb-3 text-right text-xs"><%= link_to "編集", edit_play_session_path(session), class: "text-seal" %></p><% end %>
+      <% if policy(session).update? || policy(session).destroy? %>
+        <p class="flex flex-wrap items-center justify-end gap-3 px-4 pb-3 text-xs">
```
karia: Same problem again: the `<form>` that `button_to` emits implicitly closes this `<p>`. For a GM on the session list the 削除 button ends up outside the paragraph, so `justify-end`, `px-4` and `pb-3` no longer apply to it — it renders left-aligned and flush to the card edge instead of next to 編集 on the right — plus an empty `<p></p>` is left behind.

```suggestion
        <div class="flex flex-wrap items-center justify-end gap-3 px-4 pb-3 text-xs">
```
(and change the matching `</p>` to `</div>`)

app/views/scenarios/_gallery.html.erb
karia: Fixed in ad3dee4. Confirmed with `Nokogiri::HTML5` that the `<form>` was hoisted out of the `<p>`, leaving an empty `<p></p>` behind, so this was real — switched to `<div>` here and on the other two lists.

Added a regression example ("stays in the same row as the edit link") that parses the response with the HTML5 parser and asserts the edit link is a direct sibling of the delete form. Capybara's default HTML4 parser does not reproduce the browser behaviour, so the test has to bypass it.

Also changed the scenario table's wrapper from `<span>` to `<div>`: a `<form>` is flow content and does not belong inside a phrasing element, even though it happens to parse.

app/views/people/index.html.erb
karia: Fixed in ad3dee4. Confirmed with `Nokogiri::HTML5` that the `<form>` was hoisted out of the `<p>`, leaving an empty `<p></p>` behind, so this was real — switched to `<div>` here and on the other two lists.

Added a regression example ("stays in the same row as the edit link") that parses the response with the HTML5 parser and asserts the edit link is a direct sibling of the delete form. Capybara's default HTML4 parser does not reproduce the browser behaviour, so the test has to bypass it.

Also changed the scenario table's wrapper from `<span>` to `<div>`: a `<form>` is flow content and does not belong inside a phrasing element, even though it happens to parse.

app/views/play_sessions/index.html.erb
karia: Fixed in ad3dee4. Confirmed with `Nokogiri::HTML5` that the `<form>` was hoisted out of the `<p>`, leaving an empty `<p></p>` behind, so this was real — switched to `<div>` here and on the other two lists.

Added a regression example ("stays in the same row as the edit link") that parses the response with the HTML5 parser and asserts the edit link is a direct sibling of the delete form. Capybara's default HTML4 parser does not reproduce the browser behaviour, so the test has to bypass it.

Also changed the scenario table's wrapper from `<span>` to `<div>`: a `<form>` is flow content and does not belong inside a phrasing element, even though it happens to parse.

PR #105, opened 2026-08-16
Cap each CI job with timeout-minutes

karia:
## What

Set `timeout-minutes` on every workflow job: 5 for `scan_ruby` / `scan_js` / `lint`, 10 for `test`, 15 for `image` — in `ci.yml`, and 15 for the release `image` job in `build.yml`.

## Why

In #104 the `test` job hung on the first `apt-get install libvips-tools` step and ran until the 6-hour GitHub Actions job limit cancelled it. Without `timeout-minutes`, any runner hiccup costs 6 hours.

Measured runs are test 1m15s / lint 19s / scan_ruby 23s / scan_js 11s / image 58s, so each cap sits at roughly 5-10x its measurement to absorb cold-cache spikes, with `image` at 15 min because a cache miss there means a full Docker build.

`build.yml` is covered for the same reason: its `concurrency` group is keyed on `github.ref`, so a hung tag build is never cancelled by a later push to `main`.

## Verification

- [x] `bin/rspec`
- [x] `prek run --all-files`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ permissions:
 jobs:
   scan_ruby:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
 
     steps:
       - name: Checkout code
@@ -29,6 +30,7 @@ jobs:
 
   scan_js:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
 
     steps:
       - name: Checkout code
@@ -44,6 +46,7 @@ jobs:
 
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     env:
       RUBOCOP_CACHE_ROOT: tmp/rubocop
     steps:
@@ -73,6 +76,7 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
 
     services:
       postgres:
@@ -126,6 +130,7 @@ jobs:
   # Dockerfile の破損が main に入ってから分かるのを防ぐ。push はしない。
   image:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -126,6 +130,7 @@ jobs:
   # Dockerfile の破損が main に入ってから分かるのを防ぐ。push はしない。
   image:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
```
karia, line 133: `.github/workflows/build.yml` is the one job left with the full 6-hour exposure this PR removes everywhere else — its `image` job has no `timeout-minutes`.

It runs on every push to `main` and on `v*` tags and does a full build + push to GHCR + provenance attestation (recent runs: 47s-110s), so the same hang as #104 costs 6 hours there too. It is also the release path, so the failure mode is worse than in CI: the release silently never publishes.

`concurrency.cancel-in-progress` does not cover this. It only rescues the case where a later push to the *same* ref arrives, and the group is `build-${{ github.ref }}`, so a `v*` tag run is never cancelled by a subsequent push to `main`.

Suggest adding `timeout-minutes: 15` there as well, matching this job.

(The five values in this file look right to me: the worst realistic cold-cache paths — `setup-ruby` cache eviction or a `.ruby-version` bump forcing a full `bundle install`, a rubocop cache miss, and a `type=gha` miss forcing a full Docker build — all land well inside these caps, so no false-positive risk.)

karia: 対応しました（7ef5cf0）。指摘のとおりで、`build.yml` の `image` にだけ 6 時間の露出が残っていました。`concurrency.group` が `build-${{ github.ref }}` である以上、タグ実行が `main` への push でキャンセルされないという点も確認しました。`ci.yml` の `image` と同じ `timeout-minutes: 15` を付けています。

PR #116, opened 2026-08-23
Run prek in CI instead of calling each linter directly

### LINKED ISSUE #114: Run prek in CI instead of calling each linter directly
prek hooks run only on developers' machines. CI calls rubocop, erb_lint and brakeman directly instead, so gitleaks, ghalint, check-yaml, end-of-file-fixer, trailing-whitespace and check-merge-conflict are never enforced on a PR. Workflow edits made through the web UI, or a `--no-verify` commit, reach main unchecked.

Running prek in CI closes the gap in one place, but it is not a one-liner: every hook here is `language: system` and shells out through `mise exec --`, while the workflow sets Ruby up with `ruby/setup-ruby` and has no mise.

## Work

- Add mise setup to the `lint` job and run `prek run --all-files`
- Drop the rubocop, erb_lint and brakeman steps that become duplicates
- Do the same for `config/ci.rb` so `bin/ci` and CI stay in sync

## To decide

rubocop runs with `-f github` and a `RUBOCOP_CACHE_ROOT` cache in CI only. Going through prek loses both, so either move the flag into the hook entry or keep rubocop as a separate CI step.

Came up in #113.

karia:
## What

Run `prek run --all-files` in the CI lint job and in `bin/ci`, and drop the steps it makes redundant (`erb_lint`, `brakeman`).

ghalint, check-yaml, end-of-file-fixer, trailing-whitespace and check-merge-conflict were never enforced on a PR before this. Secrets were not either, and still cannot be through the hook — see below.

## Why

Closes #114.

**mise's ruby is not used.** The hooks call ruby through `mise exec`, and mise builds ruby from source with ruby-build, which would add minutes to every run. `MISE_DISABLE_TOOLS: ruby,node` makes `mise exec` fall through to the ruby `ruby/setup-ruby` put on PATH, so the cached bundle is still what the hooks use.

**rubocop stays a separate step.** Moving `-f github` into the hook entry would print GitHub annotations on local commits too, so rubocop is excluded with `--skip` and keeps its own step. `RUBOCOP_CACHE_ROOT` is job-level env, so the cache is unaffected either way.

**gitleaks cannot run through prek here.** Its hook entry is `gitleaks git --pre-commit --redact --staged`, which only inspects the staging area. Nothing is ever staged in CI, so the hook reports zero findings no matter what the PR contains. It is skipped, and `gitleaks dir .` scans the working tree in its own step instead.

`brakeman` needs no flag change: `--exit-on-warn` and `--exit-on-error` are the default in 8.0.6, so the hook entry already fails on warnings.

## Verification

- [x] `bin/rspec` (650 examples, 0 failures)
- [x] `prek run --all-files`
- [x] CI on this branch — the `lint` job log shows all 7 hooks running plus `gitleaks dir` scanning 718 KB. 34s against a 5 minute timeout.

The secret scan was checked both ways: an unstaged file containing a `ghp_` token passes the prek hook and fails `gitleaks dir .`.

## Related

- #114


.github/workflows/ci.yml
```diff
@@ -80,11 +83,12 @@ jobs:
...
             rubocop-${{ runner.os }}-${{ env.DEPENDENCIES_HASH }}-
 
+      # rubocop だけ prek から外すのは、-f github の行アノテーションを PR に出すため。
       - name: Lint code for consistent style
         run: bin/rubocop -f github
 
-      - name: Lint ERB templates
-        run: bundle exec erb_lint --lint-all
+      - name: Run the remaining prek hooks
+        run: prek run --all-files --skip rubocop
```
karia: `prek run --all-files` does not actually run a secret scan here. The gitleaks hook from gitleaks v8.30.1 has `entry: gitleaks git --pre-commit --redact --staged --verbose` and `pass_filenames: false`, so it only inspects the *staged* diff. In CI nothing is ever staged, so gitleaks scans an empty diff and always exits 0 — a PR that adds a hardcoded secret would pass this step. If CI is meant to cover gitleaks, run it separately (e.g. `gitleaks dir .` / the gitleaks action) or add a CI-only hook alias that does not use `--staged`.

.github/workflows/ci.yml
```diff
@@ -70,6 +70,9 @@ jobs:
         with:
           bundler-cache: true
 
+      - name: Set up mise
+        uses: jdx/mise-action@3c2e0cf82a5b2e5249f0d3635a4d83d0ae861518 # v4.2.5
```
karia, line 74: The prek hook environments are not cached, and the `lint` job still has `timeout-minutes: 5`. On a cold runner `prek run --all-files` has to clone the two remote hook repos, build the python envs, and — for the gitleaks hook (`language: golang`) — download a Go toolchain and compile gitleaks. Locally `~/.cache/prek` is ~1GB, of which `tools/go` alone is 269MB. That work happens on every CI run and lands on top of `bin/rubocop` and erb_lint (which itself runs RuboCop over every ERB), so the 5-minute budget is easy to blow — an intermittent red CI that has nothing to do with the code under test. Worth caching `~/.cache/prek` (keyed on `.pre-commit-config.yaml`) the same way RuboCop's cache is handled, and/or raising the timeout.

.github/workflows/ci.yml
```diff
@@ -80,11 +83,12 @@ jobs:
           restore-keys: |
             rubocop-${{ runner.os }}-${{ env.DEPENDENCIES_HASH }}-
 
+      # rubocop だけ prek から外すのは、-f github の行アノテーションを PR に出すため。
       - name: Lint code for consistent style
         run: bin/rubocop -f github
```
karia, line 88: Brakeman used to run in `scan_ruby`, an independent job. Now it only runs as a prek hook in this step, which is placed after `bin/rubocop -f github` in the same job with no `if: always()`. Concretely: a PR with one style offense fails the RuboCop step, the job stops, and Brakeman / bundler-audit-adjacent hooks / ERB lint / ghalint never report — the security warnings stay hidden until the style fix lands. Consider `if: ${{ !cancelled() }}` on this step (or keeping the Brakeman scan in `scan_ruby`) so a formatting nit doesn't mask a security finding.

.github/workflows/ci.yml
karia: 対応しました。指摘のとおりで、実際に確認しました。同じ内容のファイルを未ステージで置くと `Detect hardcoded secrets....Passed`、`git add` してから走らせると `leaks found: 1` になります。CI ではステージが常に空なので、このフックは 0 件を返し続けていました。PR description の「gitleaks が CI で有効になる」という記述も誤りでした。

`--skip gitleaks` でフックを外し、作業ツリー全体を対象にする `gitleaks dir . --redact` を別ステップとして追加しました（`mise.toml` に gitleaks も追加）。同じ未ステージのファイルを置いた状態で exit code 1 を返すことを確認済みです。`config/ci.rb` にも同じステップを足しました。

.github/workflows/ci.yml
karia: 対応しました。`if: '!cancelled()'` を prek のステップと新しい secret scan のステップに付けました。rubocop が落ちても brakeman、erb_lint、ghalint、check-yaml の結果が出ます。

.github/workflows/ci.yml
karia: 見送ります。実測が根拠です。この PR の初回 CI で `lint` job は cold run で 57 秒（06:59:14 → 07:00:11）でした。`timeout-minutes: 5` に対して 4 分以上の余裕があります。指摘のとおり `~/.cache/prek` はローカルで 1GB 規模になりますが、CI の runner ではフック環境の構築を含めてこの時間に収まっています。キャッシュはキーの保守が要るので、実際にタイムアウトへ近づいた時点で入れるほうが妥当と判断しました。今回 secret scan のステップが増えたので、再実行後の時間も確認します。

PR #130, opened 2026-08-23
Remove the legacy UI assets and add a cross-screen audit

karia:
## What

Task 11 of the UI redesign: remove what the migration left behind, and add a cross-screen audit.

**Removed**

- The light design tokens `ink` / `paper` / `surface` / `seal` / `muted` / `rule`, and the `redacted` utility that used them.
- The legacy shared partials `_delete_button`, `_video_link`, `_alias_fields`, `_alias_row`, `_self_demotion_warning`, and `people/_alias_row`.
- `manage/scenarios/{index,_scenarios,new,edit}`, `manage/play_sessions/edit` and `manage/_nav` — none has a route or a caller. The last three are beyond the plan's list; the audit found them.
- `ManageHelper#manage_nav_class` and `ScenariosHelper#list_view_class`, which no template calls.
- The per-screen `content_for :ui_theme` switch, its `legacy` branch, and the `data-ui-theme` attribute.

**Three places the plan flagged as not a plain delete**

- `body` in the base layer now paints `--color-ui-background` / `--color-ui-text`. Every layout also sets the matching utilities, so the rule is redundant today, but it is the only thing that paints the ground for a render path that forgets them, and the ADR rules out a light theme. `spec/assets/ui_tokens_spec.rb` pins both.
- `@utility redacted` was unused and is gone. Details under Verification.
- The layout's `legacy` branch is gone along with the attribute it fed. With no light theme, an attribute that can only read `"dark"` invites future readers to treat the DOM as themeable. The specs that asserted `main[data-ui-theme="dark"]` now assert the dark ground on `body`, which is what they were standing in for.

**Fixed, found by the audit**

- Seven forms rendered `accessible_error_summary`, a leftover helper styling a white box with dark red text on the dark page, while the rest of the app used `shared/ui/_error_summary`. All seven now use the shared component and the helper is gone. Only reachable with a validation error, which is why a scan of production HTML did not show it.
- The toast bubble built its class list in `toast_controller.js` with removed tokens. It now matches the notice alert. Same reason it was missed: the element is created at runtime.
- The scenario list drew "新規登録" as a bare text link while every other list drew a primary button.
- The ordering screen drew "編集" as a bare text link while every other list drew a secondary button.
- Eight targets under the ADR's 44px: the skip link and the site logo in both layouts, the header profile link, and the purchase and recording link chips in the scenario card, table, detail and video-link components.

**Added**

`spec/system/cross_screen_audit_spec.rb` walks all 31 screens and checks what the per-screen specs cannot see: target size at 320px and 1280px, a focus indicator on every keyboard stop, the skip link starting each screen and landing focus in `main`, the account menu opening and closing from the keyboard, and the four list/detail actions all coming from the shared button component. axe and reflow stay in the per-screen specs, which already cover every screen.

## Why

Closes task 11 of [the UI redesign plan](docs/plans/2026-08-22-ui-redesign.md). New and old tokens and components have coexisted since task 1; the ADR schedules the removal for after the last migration PR.

## Verification

- [x] `bin/rspec`
- [x] `prek run --all-files`

`redacted`: grepped the whole worktree (not just `app/views` and `app/helpers`) including specs, `lib`, `config`, `db` and `public` — the only hit was its own definition. `git log -S redacted` shows the last consumer removed in ca1cc79 (task 4, the scenario detail redesign), which covers preparation notes and session history. Rebuilding `app/assets/builds/tailwind.css` after the deletion emits no `.redacted` rule and no `--color-ink`.

Chrome is not available in the authoring environment, so Chrome for Testing was fetched to run the system specs locally: 678 examples, 0 failures with `CHROME_BINARY` set, and the same run pends 13 system examples without it.

**Reported, not fixed**

- `shared/ui/_video_link` and the scenario detail's purchase list each render the raw URL as a second link to the same destination as the labelled link above it. Both now meet 44px, but the duplicate target is still a screen-reader and pointer annoyance. Removing the second link would change behaviour, so it is left for a follow-up.
- The account menu closes on an outside click but not on Escape, and focus does not return to the toggle. It is a disclosure rather than a `role="menu"` widget, so this meets WCAG 2.2 AA; it is listed only because the audit looked.

## Related

[ADR-0002](docs/adr/0002-user-interface-design.md), [UI redesign plan](docs/plans/2026-08-22-ui-redesign.md) task 11.

## Left for a follow-up

Raised in review, not addressed here:

- Deleting `manage/_nav.html.erb` orphans the `variant: :bar` branch of `shared/ui/_navigation` — the same kind of leftover this PR removes elsewhere.
- The audit spec has four gaps of its own: it exempts every `display: inline` anchor rather than only in-sentence links; its focus carve-out passes unconditionally when the document loses focus; it only ever runs signed in, so the anonymous header and `new_registration_path` go unchecked (`layouts/application.html.erb:70` draws 新規登録 as a bare underlined link); and `walk_with_tab` exits at `limit: 80` without failing, so coverage shrinks silently as a screen grows controls.


app/views/shared/ui/_video_link.html.erb
```diff
@@ -6,9 +6,9 @@
   </div>
 <% else %>
   <div class="grid gap-1">
-    <%= link_to link_text, url, target: "_blank", rel:, class: "font-bold text-ui-action" %>
+    <%= link_to link_text, url, target: "_blank", rel:, class: "inline-flex min-h-11 items-center font-bold text-ui-action" %>
```
karia: `inline-flex` on a grid item is blockified to `flex`, and a block-level flex container fills its grid area, so both links here now stretch to the full width of the card instead of hugging their text. Concretely: on the play-session detail page a recording row is one 44px-tall, full-width link, and clicking the empty space to the right of the label now opens the external URL in a new tab. Combined with the duplicate URL link you already flagged, the whole box becomes two stacked external-navigation targets. `inline-flex items-center py-…` with an explicit `min-h-11` on a non-stretching parent (or `w-fit`) keeps the 44px height without widening the hit area.

app/views/scenarios/show.html.erb
```diff
@@ -43,7 +42,7 @@
...
     <section class="mt-10"><h2 class="font-display text-xl font-bold">入手先</h2><ul class="mt-4 grid gap-3 text-sm">
-      <% @scenario.purchase_links.each do |link| %><li class="rounded-ui-control border border-ui-outline bg-ui-surface-solid p-4"><% if link.url.present? %><div class="grid gap-1"><%= link_to link.label, link.url, target: "_blank", rel: "noopener nofollow", class: "font-bold text-ui-action" %><%= link_to link.url, link.url, target: "_blank", rel: "noopener nofollow", class: "text-ui-caption text-ui-text-muted" %></div><% else %><span><%= link.label %></span><% end %></li><% end %>
+      <% @scenario.purchase_links.each do |link| %><li class="rounded-ui-control border border-ui-outline bg-ui-surface-solid p-4"><% if link.url.present? %><div class="grid gap-1"><%= link_to link.label, link.url, target: "_blank", rel: "noopener nofollow", class: "inline-flex min-h-11 items-center font-bold text-ui-action" %><%= link_to link.url, link.url, target: "_blank", rel: "noopener nofollow", class: "inline-flex min-h-11 items-center text-ui-caption text-ui-text-muted" %></div><% else %><span><%= link.label %></span><% end %></li><% end %>
```
karia: Same blockification issue as `shared/ui/_video_link`: these two links sit in a `grid gap-1`, so `inline-flex` computes to `flex` and each one stretches across the whole `<li>`. The 入手先 card now has two full-width, 44px-tall targets to the same external URL, and clicking anywhere in the card's whitespace navigates off-site — previously only the label text and the URL text were clickable.

spec/system/cross_screen_audit_spec.rb
```diff
@@ -0,0 +1,256 @@
...
+      const box = (el) => {
+        const own = el.getBoundingClientRect()
+        const label = el.closest("label")
+        if (!label) return { width: own.width, height: own.height }
+        const wrapper = label.getBoundingClientRect()
+        return { width: Math.max(own.width, wrapper.width), height: Math.max(own.height, wrapper.height) }
+      }
+      return Array.from(document.querySelectorAll(#{INTERACTIVE.dump}))
+        .filter(shown)
+        .filter((el) => !(el.tagName === "A" && getComputedStyle(el).display === "inline"))
```
karia, line 44: This exempts every anchor whose computed display is `inline`, but WCAG 2.5.8's exception is only for links inside a sentence of text. A standalone bare `link_to` has `display: inline` too, so it is silently exempted — which is exactly the defect class this audit is meant to catch. Concretely: re-adding `<%= link_to "新規登録", new_scenario_path, class: "text-sm font-bold text-ui-action" %>` to the scenario list gives a ~20px-tall target and this spec still passes. Narrowing the exemption to anchors whose parent is a text-bearing block (e.g. `el.parentElement.textContent.trim() !== el.textContent.trim()`) would keep the sentence case exempt without the blanket pass.

spec/system/cross_screen_audit_spec.rb
```diff
@@ -0,0 +1,256 @@
...
+    (() => {
+      const active = document.activeElement
+      if (!active || active === document.body || active === document.documentElement) return null
+      const style = getComputedStyle(active)
+      const outlined = style.outlineStyle !== "none" && parseFloat(style.outlineWidth) > 0
+      const ringed = style.boxShadow !== "none" && style.boxShadow !== ""
+      return {
+        name: (active.getAttribute("aria-label") || active.textContent || active.value || active.name || active.id || "").trim().slice(0, 60),
+        tag: active.tagName.toLowerCase(),
+        indicated: !active.matches(":focus") || outlined || ringed
```
karia, line 22: `!active.matches(":focus")` reports "indicated" for anything that is not matched by `:focus`, which is much broader than the date-picker case the comment describes. If the document ever loses focus while the walk runs (a stray window activation on a non-headless run, or a control that moves focus into a shadow root), every stop comes back `indicated: true` and the whole example passes without checking anything — a genuinely missing focus ring would not be reported. Scoping the escape hatch to the known case (e.g. `active.tagName === "INPUT" && active.type === "date"`) keeps the exemption without turning the assertion into a no-op.

spec/system/cross_screen_audit_spec.rb
```diff
@@ -0,0 +1,256 @@
...
+    expect(strays).to be_empty, "actions drawn outside the shared button component:\n" +
+      strays.map { |control| %(#{control['path']}: #{control['name']} as <#{control['tag']} class="#{control['classes']}">) }.join("\n")
+  end
+
+  private
+    def every_screen
+      @every_screen ||= begin
+        records = audit
+        [
+          root_path,
```
karia, line 161: Every example signs in as an admin first, so no signed-out screen is ever audited: neither `new_registration_path` nor the anonymous header. That header already violates the check three lines below — `app/views/layouts/application.html.erb:70` draws 新規登録 as `inline-flex min-h-11 items-center rounded-ui-control px-2 font-bold text-ui-text underline`, i.e. an underlined text link rather than the shared button, and its `textContent` is exactly `新規登録`. Visiting `root_path` and `new_registration_path` once while signed out would make the "shared button component" example fail today; as written the audit reports the app as clean.

spec/system/cross_screen_audit_spec.rb
```diff
@@ -0,0 +1,256 @@
+require "rails_helper"
+
+# 個別の spec は自分の画面しか見ないため、画面をまたいで初めて分かる差異はここで固定する。
+RSpec.describe "Cross-screen audit" do
+  INTERACTIVE = 'a[href], button, input:not([type="hidden"]), select, textarea, summary, [tabindex]:not([tabindex="-1"])'.freeze
```
karia: Constant assignment inside an `RSpec.describe` block resolves against the lexical scope, which is top level — so `INTERACTIVE`, `FOCUS_INDICATOR`, `SMALL_TARGETS`, `SHARED_BUTTON_SHAPE`, `ACTION_LABELS` and `NAMED_CONTROLS` all land on `Object`, visible to the whole suite and to the app under eager loading. The names are generic enough to collide: the moment a second spec (or an app class) defines `ACTION_LABELS` or `INTERACTIVE`, Ruby prints "already initialized constant" and whichever file loads last silently wins for both. `let(:…)` or lowercase locals inside the group avoid it.

spec/system/cross_screen_audit_spec.rb
```diff
@@ -0,0 +1,256 @@
...
+    end
+
+    # visit の直後に Tab を送ると、まだ前の画面が居ることがある。読み込みを待ってから焦点を先頭へ戻す。
+    def start_from_the_top_of(path)
+      expect(page).to have_css("main#main-content"), path
+      page.evaluate_script("document.activeElement && document.activeElement.blur()")
+    end
+
+    # body へ戻るか、先頭の停留点へ一周するまで辿る。
+    def walk_with_tab(path, limit: 80)
```
karia, line 245: The walk stops after 80 tabs with no assertion that the tour actually completed, so on a screen with more than 80 focus stops the remaining controls are never checked and the example still passes. The scenario list is the realistic case: the filter panel already contributes player-count buttons plus one checkbox per game system and per author, and that count grows with the data. Raising the limit does not fix it — the loop should fail (or at least report) when it exits on `limit` rather than on the wrap-around, otherwise coverage shrinks silently as screens grow.

app/views/manage/_nav.html.erb
```diff
@@ -1,6 +0,0 @@
-<% items = [
-  { label: Group.model_name.human, path: manage_groups_path, current: request.path.start_with?(manage_groups_path) },
-  { label: User.model_name.human, path: manage_users_path, current: request.path.start_with?(manage_users_path) },
-  { label: SiteSetting.model_name.human, path: manage_site_setting_path, current: request.path.start_with?(manage_site_setting_path) }
-] %>
-<%= render "shared/ui/navigation", label: "管理", items:, variant: :bar %>
```
karia, line 6: This was the only caller of `shared/ui/_navigation` with `variant: :bar`, so deleting it orphans the `:bar` branch of that partial (`app/views/shared/ui/_navigation.html.erb:7` and `:16`, plus the `bar:` entry in `variants`) — dead markup of exactly the kind this PR is removing everywhere else. `yield :sub_nav` in the application layout has no provider either, though that one predates this change.

app/views/shared/ui/_video_link.html.erb
karia: 対応しました。`grid` の子に置いた `inline-flex` が blockify されて `flex` になり、グリッド幅いっぱいのクリック領域になる件は再現条件まで確認しました。`min-h-11` と組み合わさって、行の余白をクリックしても外部サイトへ飛ぶ状態でした。`justify-self-start` を足して内容幅に留めています。

app/views/scenarios/show.html.erb
karia: 対応しました。入手先の 2 リンクも同じ `grid gap-1` の直下なので、同様に `justify-self-start` を足しています。なお同じ宛先へラベルと生 URL の 2 リンクが並ぶ点は挙動の変更になるため、PR 本文の報告のまま残します。

spec/system/cross_screen_audit_spec.rb
karia: 見送ります。指摘は正しく、`display: inline` の anchor を一律で免除するため、地の文の中のリンクではない小さな `link_to` も静かに通ります。ただし正確に判定するには「文中かどうか」を DOM から決める必要があり、今回の PR で入れるには判定の設計が要ります。監査 spec の精度向上として別途扱うほうが妥当と考えました。

spec/system/cross_screen_audit_spec.rb
karia: 見送ります。指摘のとおり `!active.matches(":focus")` は無条件パスで、ドキュメントがフォーカスを失うと何も検証しません。この carve-out は Chrome の日時入力がシャドウ DOM 側のカレンダーボタンを別のタブ停止として露出し、`:focus` が host に当たらないための回避です。条件を絞るには shadow host かどうかの判定が要るので、これも監査 spec の精度向上として別に扱います。

spec/system/cross_screen_audit_spec.rb
karia: 見送ります。監査が常にログイン済みでしか走らないため、匿名時のヘッダーと `new_registration_path` が対象外という指摘は正しいです。`layouts/application.html.erb:70` の新規登録が素の下線リンクである点も確認しました。匿名パスの追加は監査の対象範囲を広げる変更なので、別 PR で扱います。

spec/system/cross_screen_audit_spec.rb
karia: 対応しました。6 つの定数を `CrossScreenAudit` モジュール配下へ移し、`Object` への漏れを止めています。`ACTION_LABELS` や `INTERACTIVE` のような一般名が他所と衝突して警告とともに値が変わる、という指摘のとおりの risk でした。

spec/system/cross_screen_audit_spec.rb
karia: 見送ります。`limit: 80` で失敗せず抜けるため、画面のコントロール数が増えるとカバレッジが静かに縮む、という指摘は正しいです。絞り込みパネルがデータ量に比例する点も含めて妥当な懸念ですが、上限に達したら失敗させる変更は監査の落ち方を変えるので、他の spec 精度の指摘とまとめて別途扱います。

app/views/manage/_nav.html.erb
karia: 記録しておきます。`manage/_nav.html.erb` が `variant: :bar` の唯一の呼び出し元で、削除により `shared/ui/_navigation` のその分岐が孤児になる、という指摘はそのとおりです。この PR が他所で消しているのと同種の残骸なので、次の掃除の対象として PR 本文に追記します。

PR #133, opened 2026-08-24
Fix the phone header layout

karia:
## What

Drop the Google sign-in button from the header, and stop the remaining `button_to` forms from taking a whole row each.

## Why

On a phone the signed-out header wrapped onto three rows and the actions sat on the left, pushing the scenario list down the page.

Removing the Google button was the ask — it is turning into an admin-only route, and the registration page already carries it. But it was not the cause. `button_to` wraps its button in a `<form>`, and that form is a block-level flex item, so **each login control claimed a full-width row on its own** and `justify-end` could not take effect. `form: { class: "contents" }` removes the box so the button itself becomes the flex item.

Measured on the signed-out home page with a real browser:

| | 320px | 390px |
| --- | --- | --- |
| before | 3 rows, header 177px, actions from x=16 | 3 rows, header 177px, actions from x=16 |
| after | 1 row, header 125px, right edge at 304 | 1 row, header 125px, right edge at 374 |

`spec/system/application_shell_spec.rb` now pins both the row count and the right edge at those widths, so a future `button_to` in the header fails instead of silently wrapping.

## Notes

Twelve system specs signed in by clicking the header's Google button, so they now go through the registration page via a `sign_in_with_google` helper. The helper waits for the flash before navigating on — `click_button` and `visit` both return before the redirect lands, which is what `profile_and_spoiler_spec` was working around locally. That workaround moved into the helper and now covers all twelve.

`spec/requests/sign_in_button_spec.rb` keeps every Google-specific guarantee (Turbo opt-out, POST-only, origin passthrough, CSP form-action); the assertions just point at the registration page now.

`CLAUDE.md` gains the Chrome setup for `spec/system/`, plus the note that `bin/rspec` reads a stale `app/assets/builds/tailwind.css` unless you rebuild first — that bit me while measuring this change.

## Verification

- [x] `bin/rspec` — 679 examples, 0 failures, with `CHROME_BINARY` set so the system specs actually run
- [x] `prek run --all-files`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,9 +34,12 @@ mise に Chrome 本体は無いため、Playwright から取る。
 
 ```bash
 mise exec -- npx playwright install chromium
-export CHROME_BINARY=$(echo ~/.cache/ms-playwright/chromium-*/chrome-linux/chrome | tr ' ' '\n' | tail -1)
+CHROME_BINARY=$(ls -t ~/.cache/ms-playwright/chromium-*/chrome-linux/chrome 2>/dev/null | head -1)
+[ -x "$CHROME_BINARY" ] && export CHROME_BINARY || echo "Chrome が見つからない。install からやり直す"
 bin/rspec spec/system/
 ```
+
+`CHROME_BINARY` に存在しないパスを入れると、spec は skip ではなく全件 Selenium の起動失敗になる。上のように実在を確かめてから export する。
 
 **`bin/rspec` は `app/assets/builds/tailwind.css` をそのまま読む。** 手元のビルドが古いと、テンプレートを直したのに spec は前の CSS で描画する。
 クラスを足したり消したりしたら、system spec の前に `bin/rails tailwindcss:build` を流す。CI は `assets:precompile` を挟むのでこの問題は起きない。
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,6 +27,20 @@
 
 Ruby と Node は mise で固定している。shims を PATH に置いていない環境では `mise exec --` を頭に付ける。
 
+## ブラウザを使う spec
+
+`spec/system/` は Chrome が無いと skip する。見た目とアクセシビリティの確認はここにしか無いので、手元でも動かせるようにしておく。
+mise に Chrome 本体は無いため、Playwright から取る。
+
+```bash
+mise exec -- npx playwright install chromium
+export CHROME_BINARY=$(echo ~/.cache/ms-playwright/chromium-*/chrome-linux/chrome | tr ' ' '\n' | tail -1)
+bin/rspec spec/system/
+```
+
+**`bin/rspec` は `app/assets/builds/tailwind.css` をそのまま読む。** 手元のビルドが古いと、テンプレートを直したのに spec は前の CSS で描画する。
+クラスを足したり消したりしたら、system spec の前に `bin/rails tailwindcss:build` を流す。CI は `assets:precompile` を挟むのでこの問題は起きない。
+
 ## データベースを変える
 
 ```bash
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -30,14 +30,14 @@
           TRPGカタログ
         <% end %>
 
-        <nav class="flex w-full min-w-0 flex-wrap items-center justify-start gap-2 text-sm sm:ml-auto sm:w-auto sm:justify-end" aria-label="主要">
+        <nav class="flex w-full min-w-0 flex-wrap items-center justify-end gap-2 text-sm sm:ml-auto sm:w-auto" aria-label="主要">
           <% if signed_in? %>
             <% if current_person %>
               <%= link_to person_path(current_person), class: "inline-flex min-h-11 max-w-40 items-center text-ui-text-muted underline" do %><span class="min-w-0 truncate"><%= current_person.display_name %></span><% end %>
             <% else %>
               <span class="text-ui-text-muted">未登録</span>
             <% end %>
-            <%= button_to "ログアウト", session_path, method: :delete,
+            <%= button_to "ログアウト", session_path, method: :delete, form: { class: "contents" },
               class: "min-h-11 cursor-pointer rounded-ui-control px-2 text-ui-text-muted underline hover:text-ui-text" %>
             <% if current_person %>
               <div class="relative" data-controller="dropdown" data-action="click@window->dropdown#close">
@@ -69,14 +69,11 @@
           <% else %>
             <%= link_to "新規登録", new_registration_path,
               class: "inline-flex min-h-11 items-center rounded-ui-control px-2 font-bold text-ui-text underline" %>
+            <%# Google は管理者しか使わなくなったため、ヘッダーには出さず新規登録ページにだけ残す。 %>
             <%# Turbo はフォーム送信を横取りするが、provider への cross-origin リダイレクトを追えない。 %>
-            <%= button_to "Googleでログイン", "/auth/google_oauth2", method: :post,
-                  params: { origin: request.fullpath },
-                  form: { data: { turbo: false } },
-                  class: "min-h-11 cursor-pointer rounded-ui-control border border-ui-outline-strong bg-ui-surface-solid px-3 py-2 font-bold text-ui-text hover:bg-ui-subtle" %>
             <%= button_to "Discordでログイン", "/auth/discord", method: :post,
                   params: { origin: request.fullpath },
-                  form: { data: { turbo: false } },
+                  form: { class: "contents", data: { turbo: false } },
                   class: "min-h-11 cursor-pointer rounded-ui-control border border-ui-outline-strong bg-ui-surface-solid px-3 py-2 font-bold text-ui-text hover:bg-ui-subtle" %>
           <% end %>
         </nav>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -67,7 +67,7 @@
               </div>
             <% end %>
           <% else %>
-            <%= link_to "新規登録", new_registration_path,
+            <%= link_to "新規登録", new_registration_path(origin: request.fullpath),
               class: "inline-flex min-h-11 items-center rounded-ui-control px-2 font-bold text-ui-text underline" %>
             <%# Google は管理者しか使わなくなったため、ヘッダーには出さず新規登録ページにだけ残す。 %>
             <%# Turbo はフォーム送信を横取りするが、provider への cross-origin リダイレクトを追えない。 %>
--- a/spec/requests/sign_in_button_spec.rb
+++ b/spec/requests/sign_in_button_spec.rb
@@ -6,14 +6,13 @@ RSpec.describe "The sign-in button" do
 
     expect(response.body).to include(%(href="#{new_registration_path}"))
     expect(response.body).to include(">新規登録</a>")
-    expect(response.body).to include(">Googleでログイン</button>")
     expect(response.body).to include(">Discordでログイン</button>")
   end
 
   # Turbo はフォーム送信を fetch に置き換えるため、Google への cross-origin リダイレクトを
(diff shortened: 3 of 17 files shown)


app/views/layouts/application.html.erb
```diff
@@ -69,14 +69,11 @@
           <% else %>
             <%= link_to "新規登録", new_registration_path,
               class: "inline-flex min-h-11 items-center rounded-ui-control px-2 font-bold text-ui-text underline" %>
+            <%# Google は管理者しか使わなくなったため、ヘッダーには出さず新規登録ページにだけ残す。 %>
```
karia, line 72: **Google sign-in now always dumps the user back on the registration page.**

With the header button gone, `/registration/new` is the only Google entry point, and its form posts `origin: request.fullpath` — i.e. `/registration/new` itself. `SessionsController#create` redirects to `url_from(request.env["omniauth.origin"])`, and `RegistrationsController#new` has no signed-in guard, so a successful sign-in lands on a page that still reads "全機能を利用するには…ログインが必要です" with the two sign-in buttons.

Verified locally:

```
post "/auth/google_oauth2", params: { origin: new_registration_path }
follow_redirect!  # => 302 /registration/new
follow_redirect!  # body still includes "以下のステップで登録してください" and the Google form
```

Before this change the header button carried the page the admin was actually on, so this never happened for the Google flow.

Worth noting that the new `sign_in_with_google` helper papers over it — the trailing `visit root_path` (with the comment "従来どおり一覧から始められるよう戻す") means no spec observes the landing page. Consider redirecting signed-in visitors away from `registrations#new`, or falling back to `root_path` in `return_to_url` when the origin is the registration page.

spec/system/application_shell_spec.rb
```diff
@@ -33,4 +32,33 @@
...
+        })()
+      JS
+
+      expect(geometry["rows"]).to eq(1), "#{width}px で操作が #{geometry["rows"]} 段になっている"
+      expect(geometry["right"]).to be_within(1).of(width - 16),
+        "#{width}px で右端が #{geometry["right"]} にあり右寄せになっていない"
+    end
+  ensure
+    # 幅を戻さないと、同じセッションを使う後続の example が狭いままになる。
+    page.current_window.resize_to(1280, 900)
```
karia: **The `ensure` fires on the skip path and raises there.**

`skip` on line 39 raises before any driver is used, but `ensure` still runs, and without `CHROME_BINARY` the driver is `rack_test`, where `current_window` is unsupported. Reproduced:

```
2) Application shell keeps the signed-out header actions on one right-aligned row on a phone
   # Chrome is required for viewport checks
   Failure/Error: page.current_window.resize_to(1280, 900)
   Capybara::NotSupportedByDriverError:
     Capybara::Driver::Base#current_window_handle
   # ./spec/system/application_shell_spec.rb:62
   # --- Caused by: RSpec::Core::Pending::SkipDeclaredInExample
```

The suite still reports 0 failures, but the skip is no longer clean, and any real failure inside the example body would be reported with this `ensure` error stacked on top. Guarding the resize (`page.current_window.resize_to(...) if ENV["CHROME_BINARY"].present?`) or moving the skip into a `before` hook fixes it.

spec/requests/sign_in_button_spec.rb
```diff
@@ -25,19 +24,20 @@
...
   end
 
   it "passes the current URL through the sign-in flow" do
-    get root_path(order: "title_desc")
+    get new_registration_path
 
     form = response.body[%r{<form[^>]*action="/auth/google_oauth2".*?</form>}m]
 
-    expect(form).to include(%(name="origin" value="/?order=title_desc"))
+    expect(form).to include(%(name="origin"))
```
karia: **This assertion is now tautological — the origin passthrough guarantee is no longer covered.**

The PR description says the spec "keeps every Google-specific guarantee (… origin passthrough …)", but the old version fetched `root_path(order: "title_desc")` and asserted `value="/?order=title_desc"`, which is what proved `origin` captures *the page the visitor came from*. Now it fetches `/registration/new` and asserts the origin equals `/registration/new` — that passes even if `request.fullpath` were replaced with a hard-coded literal.

The one place `origin` is still dynamic is the header Discord button (`params: { origin: request.fullpath }`), and nothing asserts it any more: a change there to a fixed path would go green.

Suggest keeping a dynamic-origin assertion against the header Discord form, e.g. `get root_path(order: "title_desc")` and matching the `/auth/discord` form for `value="/?order=title_desc"`.

spec/system/application_shell_spec.rb
```diff
@@ -33,4 +32,33 @@
...
+      page.current_window.resize_to(width, 900)
+      visit root_path
+
+      geometry = page.evaluate_script(<<~JS)
+        (function () {
+          var nav = document.querySelector('header nav[aria-label="主要"]');
+          var items = [].slice.call(nav.querySelectorAll('a, button'));
+          var tops = items.map(function (e) { return Math.round(e.getBoundingClientRect().top); });
+          var rights = items.map(function (e) { return Math.round(e.getBoundingClientRect().right); });
+          return { rows: tops.filter(function (t, i) { return tops.indexOf(t) === i; }).length,
```
karia: **`rows` counts distinct `top` offsets, not rows — it will report a false "2 段" if the two controls ever differ in height.**

`nav` has `items-center`, so flex items of different heights get different `getBoundingClientRect().top` even when they sit on the same line. It works today only because the link and the button both resolve to exactly 44px (`min-h-11`). Give the Discord button a longer label that wraps, or bump one control's padding, and this fails with "2 段になっている" while the header is in fact on one row — a misleading failure that sends the next reader after the wrong bug. Comparing `Math.min(tops)` against `Math.max(bottoms)` (or checking distinct *vertical centers*) is sturdier.

Separately, line 57 compares against `width - 16`, i.e. the width requested from `resize_to`, not the viewport. `resize_to` sets the *window* rect; if the viewport ever comes back narrower (scrollbar, any browser chrome in a future Chrome/headless mode), this fails for a reason unrelated to alignment. The example above it deliberately uses `window.innerWidth` — worth returning `window.innerWidth` from the same script and asserting against that.

CLAUDE.md
```diff
@@ -27,6 +27,20 @@
...
 Ruby と Node は mise で固定している。shims を PATH に置いていない環境では `mise exec --` を頭に付ける。
 
+## ブラウザを使う spec
+
+`spec/system/` は Chrome が無いと skip する。見た目とアクセシビリティの確認はここにしか無いので、手元でも動かせるようにしておく。
+mise に Chrome 本体は無いため、Playwright から取る。
+
+```bash
+mise exec -- npx playwright install chromium
+export CHROME_BINARY=$(echo ~/.cache/ms-playwright/chromium-*/chrome-linux/chrome | tr ' ' '\n' | tail -1)
```
karia: **This can set `CHROME_BINARY` to a path that does not exist, which turns the documented "Chrome が無ければ skip" contract into hard errors.**

In bash, an unmatched glob is left literal, so if `npx playwright install chromium` was skipped or installed somewhere else, `CHROME_BINARY` becomes the literal `~/.cache/ms-playwright/chromium-*/chrome-linux/chrome`. `ENV["CHROME_BINARY"].present?` is then true, `driven_by :headless_chromium` is selected, and every system spec fails on browser launch instead of skipping.

Also, `tail -1` picks the *lexicographically* last match, not the newest build — fine for `chromium-1181` vs `chromium-1187`, wrong the moment the revision numbers change digit count.

Something like this both picks the newest and fails loudly:

```suggestion
export CHROME_BINARY=$(ls -dt ~/.cache/ms-playwright/chromium-*/chrome-linux/chrome | head -1)
```

app/views/layouts/application.html.erb
```diff
@@ -30,14 +30,14 @@
...
-        <nav class="flex w-full min-w-0 flex-wrap items-center justify-start gap-2 text-sm sm:ml-auto sm:w-auto sm:justify-end" aria-label="主要">
+        <nav class="flex w-full min-w-0 flex-wrap items-center justify-end gap-2 text-sm sm:ml-auto sm:w-auto" aria-label="主要">
           <% if signed_in? %>
             <% if current_person %>
               <%= link_to person_path(current_person), class: "inline-flex min-h-11 max-w-40 items-center text-ui-text-muted underline" do %><span class="min-w-0 truncate"><%= current_person.display_name %></span><% end %>
             <% else %>
               <span class="text-ui-text-muted">未登録</span>
             <% end %>
-            <%= button_to "ログアウト", session_path, method: :delete,
+            <%= button_to "ログアウト", session_path, method: :delete, form: { class: "contents" },
```
karia, line 40: **The signed-in half of the fix has no regression test.**

`form: { class: "contents" }` here is the same fix as on the Discord button, but the new spec (`keeps the signed-out header actions on one right-aligned row on a phone`) only visits `root_path` while signed out, so it never exercises this form. The signed-in header at 320px is arguably the tighter case — display-name link (`max-w-40 truncate`) + logout + メニュー — and dropping `contents` here would silently reintroduce the three-row wrap the PR is fixing.

The existing signed-in example above only asserts `scrollWidth <= innerWidth`, which stays true when the header wraps. Adding the signed-in state to the new row-count example (it already has a signed-in session available via `sign_in_with_google`) would pin both halves.

app/views/layouts/application.html.erb
karia: 対応しました。指摘のとおりで、ログインに成功したのに「以下のステップで登録してください」の画面へ戻される状態でした。しかも `sign_in_with_google` の末尾の `visit root_path` がそれを隠しており、spec では検知できませんでした。

新規登録ページの `origin` を `request.fullpath` から「渡された `origin`、無ければ `root_path`」に変え、ヘッダーの新規登録リンクが `origin: request.fullpath` を引き継ぐようにしました。これで元いた画面へ戻ります。`//evil.example.com/` のような値は `root_path` へ落とすようにし、その挙動も spec で固定しています。

spec/system/application_shell_spec.rb
karia: 対応しました。`skip` の後も `ensure` が走って rack_test で `Capybara::NotSupportedByDriverError` になる件は再現しました。`skip` の後に `begin`/`ensure` を置く形へ直しています。Chrome 無しで実行すると 2 examples 2 pending、エラー出力なしになることを確認しました。

spec/requests/sign_in_button_spec.rb
karia: 対応しました。`/registration/new` を GET して origin が `/registration/new` であることを確認する、というトートロジーになっていました。動的に変わるのはヘッダーの Discord ボタンと、新規登録ページが引き継ぐ origin の 2 つなので、両方を検証する形に変えています。`root_path(order: "title_desc")` から Discord フォームの origin を、`new_registration_path(origin: ...)` から Google フォームの origin を見ます。

spec/system/application_shell_spec.rb
karia: 対応しました。`top` の異なる値を数えていた件は、要素の中心線（`(top + bottom) / 2` を 10px 単位に丸め）で判定する形へ変えました。`items-center` で高さが違っても同一行と判定します。右端の比較も `window.innerWidth` を使うようにしています。

あわせてログイン済みヘッダーの検証も足しましたが、そちらは 3 項目あり 320px では正当に 2 段になるため、行数ではなく右寄せだけを固定しています。

CLAUDE.md
karia: 対応しました。glob が一致しないとリテラルが `CHROME_BINARY` に入り、skip ではなく全件 Selenium の起動失敗になる、という指摘のとおりでした。`ls -t | head -1` で更新時刻順に取り、`[ -x ... ]` で実在を確かめてから export する形にし、その理由も本文に書いています。

app/views/layouts/application.html.erb
karia: 対応しました。新しい example にログイン済み状態を足しています。実効性は確認済みで、`form: { class: "contents" }` をログアウト側から外すと spec が落ち、戻すと通ります。

